PR #1, opened 2025-07-07
feat: Add xtask workflow and GitHub Actions CI/CD

rusty1968:
- Implement cargo-xtask automation with build, test, docs, and dist tasks
- Add GitHub Actions for CI, release, documentation, and nightly builds
- Configure rust-toolchain.toml with nightly and multi-target support
- Include automated mdbook documentation generation
- Add Cargo.lock consistency checking
- Support multi-platform builds and cross-compilation


aspeeddylan:
## Description

The `book.toml` configuration file contained an invalid `multilingual = false` field that is not recognized by the current version of mdbook, causing `cargo xtask docs` to fail with a TOML parse error.

## Error

```
ERROR Invalid configuration file
        Caused by: TOML parse error at line 4, column 1
  |
4 | multilingual = false
  | ^^^^^^^^^^^^
unknown field `multilingual`, expected one of `title`, `authors`, `description`, `src`, `language`, `text-direction`
```

## Solution

Removed the `multilingual = false` field from:
- `docs/book.toml` - the actual configuration file
- `xtask/src/docs.rs` - the template that generates default configurations

## Result

The `cargo xtask docs` command now executes successfully and builds the documentation without errors.

PR #16, opened 2025-08-01
I2C target traits

rusty1968:
Traits modeling I2C target behavior.


stevenlee0809 (commented):
@rusty1968 
Looks good to me, I don't have permission to approve, so just leaving a comment here.

PR #21, opened 2025-09-03
Initial mctp baremetal implementation

embediver:
Implementation of a baremetal mctp router that uses the available types from [mctp-estack]().

The router currently only supports a single transport binding (so no real routing/bridge functionality).
The `mctp-rs` blocking traits for requests and listeners are not implemented right now since that is difficult without any runtime.


embediver:
Closing this in favor of https://github.com/OpenPRoT/mctp-lib/pull/3.

PR #70, opened 2025-10-15
Reviewed and Updated the Threat Modeling

bikashpoudel43:
Summary of changes

- Converted the asset listing into a table
- Added short description for attacks
- Added an example of threat modeling as per the spec


docs/src/specification/threat_model.md
```diff
@@ -1,37 +1,43 @@
 # Threat Model
 
-## Assets
+## Assets and Desired Security Properties
```
bikashpoudel43: I thought it would be better to organize this as table so that we could have assets listed in the first column.

docs/src/specification/threat_model.md
```diff
@@ -1,37 +1,43 @@
...
--   Integrity and authorization of cryptographic operations
--   Integrity of anti-rollback counters
--   Integrity and confidentiality of symmetric keys managed by OpenPRoT
--   Integrity and confidentiality of private asymmetric keys
--   Integrity of boot measurements
--   Integrity and authenticity of firmware update payloads
--   Integrity and authenticity of OpenPRoT policies
+| Asset                                  | Security Property         | Description |
+|----------------------------------------|---------------------------|-------------|
+| OpenPRoT Firmware                      | Integrity, Authenticity, Non-Repudiation | Ensure firmware is unmodified, from a trusted source, and the origin cannot de denied.|
```
bikashpoudel43: Why non-repudiation?
I think we have to consider the supply chain security where a trusted source on top of signing has to provide a proof that it has done its security due diligence on their side which could be a signed audit trail.

docs/src/specification/threat_model.md
```diff
@@ -56,3 +62,50 @@ Integrators should use the JIL specification as a guideline to identify relevant
...
+    %% Class definitions for coloring
+    classDef manu fill:#cce6ff,stroke:#5588bb,stroke-width:2px;
+    classDef dev fill:#ccffcc,stroke:#55bb77,stroke-width:2px;
+
+    %% Assign classes to nodes in each subgraph
+    class MPS manu;
+    class JTAGP,JTAG,OTP dev;
+```
+
+#### Example Threat Modeling Table as Per this Specification
```
bikashpoudel43: This is not a complete list but an illustration. Which debug port or IO is used for programming depends from vendor to vendor.

PR #111, opened 2026-02-25
ci: use Bazel for GitHub Pages documentation build

moidx:
This PR transitions the documentation build process from 'cargo xtask' to Bazel in the GitHub Pages workflow. It also removes the obsolete 'xtask docs' implementation.

### DIFF
--- a/.github/workflows/github_pages.yml
+++ b/.github/workflows/github_pages.yml
@@ -34,7 +34,10 @@ jobs:
         repository-cache: true
 
     - name: Build mdbook documentation
-      run: bazel build //docs
+      id: build-docs
+      run: |
+        bazel build //docs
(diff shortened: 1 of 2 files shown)


.github/workflows/github_pages.yml
```diff
@@ -26,35 +26,23 @@ jobs:
...
+      run: bazel build //docs
 
     - name: Setup Pages
       uses: actions/configure-pages@v5
 
     - name: Upload artifact
       uses: actions/upload-pages-artifact@v4
       with:
-        path: ./target/book
+        path: ./bazel-bin/docs/docs
```
cfrantz: This should work, but we might want to use bazel to query the output provider.

moidx: Done.

Issue #115, opened 2026-02-27
Execute heavyweight tests in the nightly workflow


cfrantz:
Per the OpenPRoT WG meeting on 2026-02-27, we agreed to execute heavyweight tests such as Earlgrey verilator tests into the nightly workflow.

PR #118, opened 2026-02-27
Reconfigure `nightly` workflow

cfrantz:
Reconfigure the nightly workflow to execute all tests and build all targets.

Addresses: #115

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -12,76 +12,46 @@ env:
   RUST_BACKTRACE: 1
 
 jobs:
+  nightly-build:
+    name: Nightly Build
+    runs-on: ubuntu-latest
+
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly.yml
```diff
@@ -17,71 +17,34 @@ jobs:
...
-    - name: Check for outdated dependencies
-      run: cargo outdated --exit-code 1
-      continue-on-error: true
-    
-    - name: Install cargo-audit
-      run: cargo install cargo-audit
-    
-    - name: Run security audit
-      run: cargo audit
+    - name: Run tests
```
moidx: nit: should build first and then test? OK if you want to leave it as-is.

cfrantz: Done

PR #119, opened 2026-02-27
Remove cargo and xtask resources 

cfrantz:
The codebase has been converted to use bazel as the build system.
Remove no-longer-needed cargo resources and xtasks.

This PR depends on #116, #117, #118.


cfrantz:
Note: also updates `.gitignore` to remove cargo-specific output directories and add bazel-specific directories.

PR #160, opened 2026-04-19
earlgrey: Implement GPIO and Pinmux drivers

moidx:
- Implemented EarlGrey GPIO driver with HAL trait support
- Implemented EarlGrey Pinmux driver for pad routing
- Optimized for hardware masked registers and atomic updates
- Added Verilator smoke test for functional verification

### DIFF
--- a/target/earlgrey/drivers/gpio.rs
+++ b/target/earlgrey/drivers/gpio.rs
@@ -1,0 +1,318 @@
+// Licensed under the Apache-2.0 license
+// SPDX-License-Identifier: Apache-2.0
+
+#![no_std]
+
+use earlgrey_pinmux::{EarlGreyPinmux, Pad, PadConfig, Pull};
+use core::fmt::Debug;
+use openprot_hal_blocking::gpio_port::{
+    EdgeSensitivity, GpioError, GpioErrorKind, GpioErrorType, GpioInterrupt, GpioPort,
+    InterruptOperation, PinMask,
+};
+use registers::gpio;
+
+#[derive(Debug, Copy, Clone, Eq, PartialEq)]
+pub enum EarlGreyGpioError {
+    HardwareFailure,
+    InvalidConfiguration,
+}
+
+// TODO: figure out what we're doing with precise errors.
+impl GpioError for EarlGreyGpioError {
+    fn kind(&self) -> GpioErrorKind {
+        match self {
+            EarlGreyGpioError::HardwareFailure => GpioErrorKind::HardwareFailure,
+            EarlGreyGpioError::InvalidConfiguration => GpioErrorKind::UnsupportedConfiguration,
+        }
+    }
+}
+
+pub struct EarlGreyGpio {
+    registers: gpio::RegisterBlock<ureg::RealMmioMut<'static>>,
+    pinmux: EarlGreyPinmux,
+}
+
+impl EarlGreyGpio {
+    /// Create a new instance of the EarlGrey GPIO driver using real MMIO.
+    /// 
+    /// # Safety
+    /// 
+    /// The caller must ensure that they have exclusive access to the GPIO and Pinmux peripherals.
+    pub unsafe fn new() -> Self {
+        Self {
+            registers: unsafe { gpio::RegisterBlock::new(gpio::Gpio::PTR) },
+            pinmux: unsafe { EarlGreyPinmux::new() },
+        }
+    }
+
+    /// Read current state of output pins.
+    /// 
+    /// This is a target-specific extension not yet in the core HAL.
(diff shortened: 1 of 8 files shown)


target/earlgrey/drivers/gpio.rs
cfrantz, line 1: Lets put this in //target/earlgrey/drivers.  The current USB driver PR would have created this subdir.

target/earlgrey/drivers/pinmux.rs
cfrantz, line 1: Lets put this in //target/earlgrey/drivers.

target/earlgrey/drivers/pinmux.rs
```diff
@@ -0,0 +1,123 @@
+// Licensed under the Apache-2.0 license
+// SPDX-License-Identifier: Apache-2.0
+
+#![no_std]
+
+use registers::pinmux;
+
+#[derive(Clone, Copy, Debug, Eq, PartialEq)]
```
cfrantz, line 8: We probably want this to be `#[repr(u32)]` to force the compiler to use `u32` as the underlying representation.

We _might_ want to implement `From<u32> for Pad` and `From<Pad> for u32` to facilitate conversions (but maybe we should wait and see if we need those conversions).

target/earlgrey/drivers/gpio.rs
moidx: Done.

target/earlgrey/drivers/pinmux.rs
moidx: Done.

target/earlgrey/drivers/pinmux.rs
moidx: Let's wait and see. We can add support for this once we have an implementation plan for the component in charge of managing the platform level IO.

target/earlgrey/drivers/pinmux.rs
```diff
@@ -0,0 +1,123 @@
...
+}
+
+impl Pad {
+    pub fn is_dio(&self) -> bool {
+        (*self as u32) >= 47
+    }
+
+    pub fn dio_index(&self) -> usize {
+        (*self as usize) - 47
+    }
```
cfrantz: This can fail if `dio_index` is called on an MIO pad.

A bit more rust-idiomatic:
```
pub fn dio_index(self) -> Option<usize> {
  let index = self as usize;
  if index >= 47 {
    Some(index - 47)
  } else {
    None
  }
}
```

target/earlgrey/drivers/pinmux.rs
```diff
@@ -0,0 +1,123 @@
...
+        Self {
+            pull: Pull::None,
+            open_drain: false,
+            invert: false,
+        }
+    }
+}
+
+pub struct EarlGreyPinmux {
+    registers: pinmux::RegisterBlock<ureg::RealMmioMut<'static>>,
```
cfrantz, line 63: You _could_ keep an instance of `pinmux::PinmuxAon` here as it represents the peripheral and is fixed at the target address.  However, you'd have to  manually get access to the register block when you wanted to use it via the `regs()` or `regs_mut()` functions.

Doing it this way is nicer in that you don't have to use `regs_mut` and that you can control the peripheral address, which is certainly more convenient for peripherals that have multiple instances (like uarts, i2c and spi).

target/earlgrey/drivers/gpio.rs
```diff
@@ -0,0 +1,317 @@
...
+}
+
+impl GpioError for EarlGreyGpioError {
+    fn kind(&self) -> GpioErrorKind {
+        match self {
+            EarlGreyGpioError::HardwareFailure => GpioErrorKind::HardwareFailure,
+            EarlGreyGpioError::InvalidConfiguration => GpioErrorKind::UnsupportedConfiguration,
+        }
+    }
+}
```
cfrantz, line 28: TODO: figure out what we're doing with precise errors.

cfrantz (approved):

target/earlgrey/drivers/pinmux.rs
moidx: Fixed.

target/earlgrey/drivers/pinmux.rs
moidx: Added `repr(u32)`

target/earlgrey/drivers/gpio.rs
moidx: Added TODO

target/earlgrey/drivers/pinmux.rs
moidx: ack.

PR #168, opened 2026-04-22
util: add a simple ufmt based console 

cfrantz:


util/console/dbg_print.c
```diff
@@ -0,0 +1,131 @@
...
+                        *buf++ = (kHexTable[ch & 15]);
+                    }
+                }
+                break;
+            }
+            case 's': {
+                // Print a null-terminated string.
+                const char *str = va_arg(args, const char *);
+                while (*str != '\0') {
+                    *buf++ = (*str++);
```
moidx, line 82: nit: the function name already indicates that this function is only acceptable for debug use cases, but should we add a bound check just in case someone ends up misusing the library? 

```c
if (buf >= buffer + sizeof(buffer) - 1) break;
```

moidx (approved):

util/console/dbg_print.c
cfrantz: Fixed.

PR #177, opened 2026-04-28
Usart: user space driver crates, docs

rusty1968:
Add USART user-space driver, integration guides, and ast10x0 reference platform implementation


CourtneyDrant (approved):
Looks modular and contained.

PR #225, opened 2026-05-13
Introducing the ast10x0 board crate

rusty1968:
This PR adds the ast10x0 i2c peripheral driver and a coordinating board crate which does pre-kernel platform initialization.


target/ast10x0/harness/test_runner.py
JesseMelon, line 1: This workaround regenerates a .bin file which is already made as part of bazel's build process. But strangely, only the elf version is exposed as part of the target's runfiles. 

a patch like this: 

```
diff --git a/MODULE.bazel b/MODULE.bazel
index f2d22c8..6c76b7f 100644
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -31,6 +31,8 @@ git_override(
         "//third_party/pigweed:integration_tests_armv7m.patch",
         # Fix lost-wakeup race in ipc/user test USER signal round-trip.
         "//third_party/pigweed:ipc_user_signal_race.patch",
+        # Stage .bin alongside .elf in system_image_test runfiles.
+        "//third_party/pigweed:system_image_test_bin_runfiles.patch",
     ],
     remote = "https://pigweed.googlesource.com/pigweed/pigweed",
 )
diff --git a/third_party/pigweed/system_image_test_bin_runfiles.patch b/third_party/pigweed/system_image_test_bin_runfiles.patch
new file mode 100644
index 0000000..52b0aa9
--- /dev/null
+++ b/third_party/pigweed/system_image_test_bin_runfiles.patch
@@ -0,0 +1,8 @@
+--- a/pw_kernel/tooling/system_image.bzl
++++ b/pw_kernel/tooling/system_image.bzl
+@@ -182 +182,4 @@
+-            runfiles = ctx.attr.image[DefaultInfo].default_runfiles,
++            runfiles = ctx.runfiles(files = [
++                ctx.attr.image[SystemImageInfo].elf,
++                ctx.attr.image[SystemImageInfo].bin,
++            ]),
```
Would make the workaround unnecessary. I wonder why the current behavior omits the generated bin from the runfiles, but includes the .elf?

JesseMelon (approved):

PR #236, opened 2026-05-21
Introduce flash traits, client and service

cfrantz:
Introduce flash traits, client and server.
- `//hal/blocking/flash` contains trait abstractions for accessing flash.
- `//services/flash` contains the flash IPC client and server implementations and user docs in a README.md
- `//target/earlgrey/drivers/eflash_driver.rs` is the implementation for earlgrey's internal flash block.

I've changed `FlashAddress` to be a 2 x `u32` type that includes a `device_id` and an `offset`.  The type implements arithmetic ops on the offset.  The `device_id` can be used to encode both @rusty1968 's idea of a "routing key" to represent  a specific device or Earlgrey's notion of separate DATA and INFO address spaces (or both).  The `device_id` needs a bit more development as the current flash service doesn't look at it, but we might want it to examine it.

I've adjusted my (actually kor's) original traits to adopt @rusty1968 's notion of flash geometry.

There are several preparatory commits before the HALs and implementations.  We can also discuss the content of those commits as well.

### DIFF
--- a/hal/blocking/flash/driver.rs
+++ b/hal/blocking/flash/driver.rs
@@ -1,0 +1,131 @@
+//! Low-level flash driver interface.
+
+#![no_std]
+
+use core::num::NonZero;
+
+use util_error::ErrorCode;
+use util_types::PowerOf2Usize;
+use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
+
+/// Low-level flash driver interface.
+///
+/// This trait defines the interface for interacting with flash hardware at a low level.
+/// It supports asynchronous-style operations (start/is_busy/complete) but can also be
+/// implemented for synchronous drivers.
+pub trait FlashDriver {
+    /// A bitmap of supported erase block sizes.
+    ///
+    /// Each bit i represents a supported erase block size of 2^i bytes.
+    const ERASABLE_SIZES_BITMAP: u32;
+    /// The maximum size of a single program operation.
+    const PROGRAM_WINDOW_SIZE: usize;
+    /// The maximum size of a single read operation.
+    const MAX_READ_SIZE: usize;
+    /// The alignment required for read operations.
+    const READ_ALIGNMENT: usize;
+    /// The alignment required for program operations.
+    const PROGRAM_ALIGNMENT: usize;
+
+    /// Returns the total size of the flash in bytes.
+    fn size(&self) -> NonZero<usize>;
+
+    /// Reads data from flash.
+    ///
+    /// # Arguments
+    /// * `start_addr`: The address to start reading from.
+    /// * `buf`: The buffer to read data into.
+    fn read(&mut self, start_addr: FlashAddress, buf: &mut [u8]) -> Result<(), ErrorCode>;
+
+    /// Starts an erase operation.
+    ///
+    /// # Arguments
+    /// * `start_addr`: The start address of the block to erase.
+    /// * `size`: The size of the block to erase.
+    fn start_erase(&mut self, start_addr: FlashAddress, size: PowerOf2Usize) -> Result<(), ErrorCode>;
+
+    /// Starts a program operation.
+    ///
+    /// # Arguments
+    /// * `start_address`: The address to start programming at.
+    /// * `data`: The data to program.
+    fn start_program(&mut self, start_address: FlashAddress, data: &[u8]) -> Result<(), ErrorCode>;
+
+    /// Returns whether the driver is currently busy with an operation.
+    fn is_busy(&mut self) -> bool;
+
+    /// Completes a pending operation and returns the result.
+    fn complete_op(&mut self) -> Result<(), ErrorCode>;
+}
+
+/// Represents an address in flash memory.
+///
+/// A flash address consists of a device identifier and an offset within that
+/// device's address space.
+#[derive(Default, Clone, Copy, PartialEq, Eq, IntoBytes, Immutable, FromBytes, KnownLayout)]
+#[repr(C)]
+pub struct FlashAddress {
+    device_id: u32,
+    offset: u32,
+}
+
+impl core::fmt::Display for FlashAddress {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        write!(f, "{}:0x{:08x}", self.device_id, self.offset)
+    }
+}
+
+impl core::fmt::Debug for FlashAddress {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        core::fmt::Display::fmt(self, f)
+    }
+}
+
+impl FlashAddress {
+    /// Creates a new `FlashAddress`.
+    pub const fn new(device_id: u32, offset: u32) -> Self {
+        Self { device_id, offset }
+    }
+
+    /// Returns the device identifier.
+    pub fn device_id(&self) -> u32 {
+        self.device_id
+    }
+
+    /// Returns the offset within the device's address space.
+    pub fn offset(&self) -> u32 {
+        self.offset
+    }
+}
+
+impl core::ops::Add<usize> for FlashAddress {
+    type Output = Self;
+    fn add(self, other: usize) -> Self {
+        Self {
+            device_id: self.device_id,
+            offset: self.offset + other as u32,
+        }
+    }
+}
+
+impl core::ops::AddAssign<usize> for FlashAddress {
+    fn add_assign(&mut self, other: usize) {
+        self.offset += other as u32;
+    }
+}
+
+impl core::ops::BitAnd<usize> for FlashAddress {
+    type Output = Self;
+    fn bitand(self, other: usize) -> Self {
+        Self {
+            device_id: self.device_id,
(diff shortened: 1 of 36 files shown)


util/ipc/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+    /// * `deadline`: The time by which the transaction must complete.
+    pub fn transaction<const N: usize>(
+        &self,
+        request: &[&[u8]],
+        response: &mut [&mut [u8]],
+        deadline: Instant,
+    ) -> Result<usize, ErrorCode> {
+        if false {
+            //let _n = N;
+            //syscall::channel_transact_iovec(self.0, request, response, deadline)
```
rusty1968, line 51: nit: Delete commented out code.

rusty1968:
There are too many pieces introduced in a single PR. Maxiimum benefit can be achieved by introducing flash traits, error crate and other impactful pieces in smaller chunks.

cfrantz:
I did not explain myself well in the description: this is not meant to be submittable as-is, or even reviewed in a serious line-by-line way.  This is a working example of the flash traits, client, server and low-level drivers on earlgrey.  It's built in a way that I think should be compatible with other system and meets some of the requirements from our last discussion.

This is meant to be able to look at the big picture of how the flash traits fit together with the other pieces and to have a discussion.

I will, of course, send smaller PRs when we've worked out the high-level details.

rusty1968:
> I did not explain myself well in the description: this is not meant to be submittable as-is, or even reviewed in a serious line-by-line way. This is a working example of the flash traits, client, server and low-level drivers on earlgrey. It's built in a way that I think should be compatible with other system and meets some of the requirements from our last discussion.
> 
> This is meant to be able to look at the big picture of how the flash traits fit together with the other pieces and to have a discussion.
> 
> I will, of course, send smaller PRs when we've worked out the high-level details.

Excellent!

services/flash/server.rs
```diff
@@ -0,0 +1,112 @@
...
+    fn handle_program<'a>(&mut self, data: &'a mut [u8]) -> Result<&'a [u8], ErrorCode> {
+        let (addr, data) =
+            FlashAddress::mut_from_prefix(data).map_err(|_| error::IPC_ERROR_BAD_REQ_LEN)?;
+        self.flash.program(*addr, data)?;
+        Ok(&data[0..0])
+    }
+
+    fn handle_read<'a>(&mut self, data: &'a mut [u8]) -> Result<&'a [u8], ErrorCode> {
+        let addr =
+            FlashAddress::read_from_bytes(&data[0..8]).map_err(|_| error::IPC_ERROR_BAD_REQ_LEN)?;
```
rusty1968, line 56:
```suggestion
           FlashAddress::read_from_bytes(
    data.get(..8).ok_or(error::IPC_ERROR_BAD_REQ_LEN)?
).map_err(|_| error::IPC_ERROR_BAD_REQ_LEN)?;
```

rusty1968:
One of the aspects that Hubris did well and we kept it when porting our stacks to PW was to keep the request/response wire protocol format (encoded using IDL or done manually) from the IPC transport mechanism itself.  We would like to see the same in the flash stack. This allowed host-based tests to be performed.

hal/blocking/flash/flash.rs
```diff
@@ -0,0 +1,356 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! High-level blocking flash interface.
+
+#![cfg_attr(not(test), no_std)]
+
+use core::{cmp::min, num::NonZero};
+pub use hal_flash_driver::FlashAddress;
+use hal_flash_driver::FlashDriver;
+use util_error::ErrorCode;
```
rusty1968: hal/blocking` should be kept as a pure set of seams — no Pigweed, no OS
imports. Pigweed is being brought in by the way of `util_error`  which depends on `@pigweed//pw_status/rust:pw_status.

hal/blocking/flash/driver.rs
```diff
@@ -0,0 +1,131 @@
...
+    fn complete_op(&mut self) -> Result<(), ErrorCode>;
+}
+
+/// Represents an address in flash memory.
+///
+/// A flash address consists of a device identifier and an offset within that
+/// device's address space.
+#[derive(Default, Clone, Copy, PartialEq, Eq, IntoBytes, Immutable, FromBytes, KnownLayout)]
+#[repr(C)]
+pub struct FlashAddress {
```
rusty1968, line 75: ## The routing key concept should leverage the PW capability model
The flash service protocol carries a `device_id field in every address, implying
multi-device support.  

It follows that routing key is client-supplied — it lives inside the message payload.
A client can set `device_id` to any value.  Nothing in the current design prevents a task
that should only access one flash device from addressing another.  Enforcement would have to be a software check inside the server, applied after the message is already received.

Nothing stops one task from accessing another task's flash device. The kernel's capability model is bypassed at the device-selection boundary.
Device-level access control is delegated to software rather than enforced by the kernel.

util/ipc/lib.rs
```diff
@@ -0,0 +1,106 @@
...
+    ///
+    /// This method combines multiple request buffers into a single buffer of
+    /// size `N`, sends it over the channel, and then distributes the response
+    /// into the provided response buffers.
+    ///
+    /// # Arguments
+    /// * `request`: A slice of buffers containing the request data.
+    /// * `response`: A slice of mutable buffers to receive the response data.
+    /// * `deadline`: The time by which the transaction must complete.
+    pub fn transaction<const N: usize>(
```
rusty1968, line 43: replace transaction < const N: usize  >  with caller-supplied buffer

    /// * `buffer`: Caller-supplied scratch buffer; must be large enough to hold
    ///   the gathered request **and** the full response.
    pub fn transaction(
        &self,
        request: &[&[u8]],
        response: &mut [&mut [u8]],
        buffer: &mut [u8],
        deadline: Instant,
    ) -> Result<usize, ErrorCode> {

rusty1968:
Right now the flash request/response protocol exists only as convention — the client writes bytes at offsets it chose, and the server reads bytes at offsets it also chose, and they happen to agree. There is no single place that says "a flash IPC message is a 12-byte header followed by an operation-specific payload." If they ever diverge, the compiler won't notice.

```rust
// services/flash/client.rs
self.ipc.transaction::<16>(
    &[
        IPC_OP_FLASH_ERASE.as_bytes(),   // bytes [0..4]   opcode
        start_addr.as_bytes(),           // bytes [4..12]  FlashAddress (8 bytes)
        size_val.as_bytes(),             // bytes [12..16] u32
    ],
    &mut [result.as_mut_bytes()],
    Instant::MAX,
)?;
```
```rust
// services/flash/server.rs
// `data` arrives here with the opcode already stripped — offset accounting
// is split across handle_one() and handle_erase().
fn handle_erase<'a>(&mut self, data: &'a mut [u8]) -> Result<&'a [u8], ErrorCode> {
    let (addr, data) = FlashAddress::read_from_prefix(data)   // bytes [0..8]
        .map_err(|_| error::IPC_ERROR_BAD_REQ_LEN)?;
    let (size, _)   = u32::read_from_prefix(data)             // bytes [8..12]
        .map_err(|_| error::IPC_ERROR_BAD_REQ_LEN)?;
    ...
}
```

The contract — `FlashAddress` occupies bytes 4–12, `size` occupies bytes 12–16
— lives entirely in the protocol designer's head, duplicated across two files.

## The fix: make the struct the contract

Extract a shared `FlashEraseRequest` in a crate both sides depend on:

```rust
// services/flash/wire/lib.rs
#[derive(FromBytes, Immutable, IntoBytes, KnownLayout)]
#[repr(C)]
pub struct FlashEraseRequest {
    pub addr: WireAddress,   // 8 bytes
    pub size: u32,           // 4 bytes
}
```

The client encodes by value:

```rust
let req = FlashEraseRequest { addr: addr.into(), size: size_val };
// req.as_bytes() produces exactly the same 12 bytes — but derived from
// the struct layout, not from manually stacked slices.
```

The server decodes symmetrically:

```rust
let (req, _) = FlashEraseRequest::read_from_prefix(data)
    .map_err(|_| error::IPC_ERROR_BAD_REQ_LEN)?;
// req.addr and req.size are now typed fields, not offset arithmetic.
```

rusty1968:
The core `Flash` trait definition is well-designed — a clean, minimal interface with the right operations at the right level of abstraction.The problem is not the trait.  The problem is that the wire protocol has no room to grow beyond it.

The opcode space, the server dispatch, and the buffer constants are all built around exactly the four operations `Flash` defines today.  Adding a new operation, e.g., exposing `Flash + FlashExOp` over a single IPC handle, or versioning the protocol independently of the trait all require changes that the current design has no natural place for.

Consider `FlashExOp` — a supertrait that adds vendor-specific extended operations on top of `Flash`:

```rust
pub trait FlashExOp: Flash {
    fn ex_op(&mut self, code: u16, input: &[u8], output: &mut [u8]) -> Result<(), ErrorCode>;
}
```

Generic code that needs extended operations simply bounds on `F: Flash + FlashExOp`. The trait design is clean.  The problem is what happens when you try to expose `FlashExOp` over IPC.

PR #239, opened 2026-05-27
bazel: Fix dependabot config

cfrantz:


cfrantz:
I'm getting conflicting advice from Gemini on valid bazel dependabot configuration.  Lets let presubmit run on this and see if it complains about anything.

moidx (approved):

PR #240, opened 2026-05-28
[earlgrey] add QEMU support

moidx:
Wire the lowRISC QEMU ot-earlgrey machine into target/earlgrey as a target (parallel to verilator). QEMU runs ~50x faster than verilator (2.6s vs minutes) and is part of the earlgrey_qemu_tests workflow within ci so the tests run on every PR.

Active QEMU tests:
- //target/earlgrey/tests/ipc/user:ipc_runner_qemu_test        PASS 2.6s
- //target/earlgrey/tests/threads/kernel:threads_runner_qemu_test  PASS 12.5s
50/50 stability runs, zero flakes.

Build changes:
- third_party/qemu/: lowRISC QEMU release archive pinned to v10.2.0-2026-01-15; hermetic pip-parse for hjson + pyelftools (@openprot_python_deps, Bazel-managed Python 3.11).
- third_party/opentitan_rtl/: vendored RTL constants for cfggen. RMA OTP vmem consumed from @opentitan_devbundle (no local copy).
- target/earlgrey/tooling/qemu.bzl: qemu_cfg / qemu_otp / qemu_flash rules + pure-function helpers callable from opentitan_runner.bzl.
- target/earlgrey/tooling/qemu_runner.py + qemu_start.sh: Python harness with HMP cont, UART Unix-socket reader, pw_tokenizer detokenization, SIGTERM-safe cleanup.

Target glue:
- target_type=qemu added; entry.rs .fake_rom_ext cfg widened from `verilator` to `any(verilator, qemu)`. Silicon ROM delta verified at 0 bytes.
- config.rs gets qemu SYSTEM_CLOCK_HZ = 24_000_000 (matches opentitan/sw/device/lib/arch/device_sim_qemu.c).
- opentitan_runner.bzl gains interface="qemu" branch with _QEMU_ATTRS separated from _BASE_ATTRS to avoid testonly contamination of non-test opentitan_runner targets. interface attr gained values=[...] constraint.

CI: new earlgrey_qemu_tests workflow added to workflows.json, wired into ci group; ci_tests filters now exclude -qemu to prevent double-counting.


target/earlgrey/tooling/qemu_constants_dumper.py
cfrantz, line 1: Would this file be better located under //target/earlgrey/... ?

target/earlgrey/tooling/qemu_runner.py
```diff
@@ -0,0 +1,416 @@
...
+    try:
+        while not done.is_set():
+            try:
+                chunk = sock.recv(4096)
+            except OSError:
+                break
+            if not chunk:
+                break
+
+            line_buf += chunk.decode("latin-1")
```
cfrantz: Can we treat the line-buf as bytes here?  There's a utf-8 decode in `_emit_list`.

cfrantz (approved):
Fast review: mostly LGTM.

I think we should work towards qemu hosted under opentitantool to eliminate having two different runners.  We should also work towards enabling `opepntitanlib::test_lib` for writing host-based test harnesses, but that is beyond the scope of this PR.

target/earlgrey/tooling/qemu_constants_dumper.py
moidx: Fixed.

target/earlgrey/tooling/qemu_runner.py
moidx: Fixed.

moidx:
Agreed that we should enable a single exec environment. We should sync with lowRISC to file issues for the work required in ottool lib to support qemu.

PR #242, opened 2026-05-29
automation: Synchronize main to earlgrey-hwe

cfrantz:
1. Added a github app to manage branch synchronization actions.
2. Perform synchronization from `main` to `earlgrey-hwe`.

### DIFF
--- a/.github/workflows/sync-main-to-earlgrey-hwe.yml
+++ b/.github/workflows/sync-main-to-earlgrey-hwe.yml
@@ -1,0 +1,38 @@
+name: Scheduled Sync Main to Earlgrey HWE
+
+on:
+  schedule:
+    # Runs every day at 15:00 UTC (8:00 AM PDT / 7:00 AM PST)
+    - cron: '0 15 * * *'
+  workflow_dispatch: # Allows you to trigger it manually from the Actions tab
+
+jobs:
+  create-pull-request:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0 # Fetches all history so branches can be compared
+
+      - name: Generate GitHub App Token
+        id: generate-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.SYNCHRONIZE_BOT_APP_ID }}
+          private-key: ${{ secrets.SYNCHRONIZE_BOT_PRIVATE_KEY }}
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v6
+        with:
+          token: ${{ steps.generate-token.outputs.token }}
+          branch: automation/sync-main-to-earlgrey-hwe
+          base: earlgrey-hwe
+          title: 'chore: sync main into earlgrey hwe'
+          reviewers: cfrantz, moidx
+          body: |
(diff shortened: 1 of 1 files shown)


.github/workflows/sync-main-to-earlgrey-hwe.yml
```diff
@@ -0,0 +1,38 @@
...
+
+on:
+  schedule:
+    # Runs every day at 15:00 UTC (8:00 AM PDT / 7:00 AM PST)
+    - cron: '0 15 * * *'
+  workflow_dispatch: # Allows you to trigger it manually from the Actions tab
+
+jobs:
+  create-pull-request:
+    runs-on: ubuntu-latest
```
moidx, line 11: nit: should we pin this to an LTS release?

.github/workflows/sync-main-to-earlgrey-hwe.yml
```diff
@@ -0,0 +1,38 @@
...
+
+      - name: Generate GitHub App Token
+        id: generate-token
+        uses: actions/create-github-app-token@v1
+        with:
+          app-id: ${{ secrets.SYNCHRONIZE_BOT_APP_ID }}
+          private-key: ${{ secrets.SYNCHRONIZE_BOT_PRIVATE_KEY }}
+
+      - name: Create Pull Request
+        uses: peter-evans/create-pull-request@v6
```
moidx, line 26: commentary: mind twister. Marvel multiverse.

.github/workflows/sync-main-to-earlgrey-hwe.yml
```diff
@@ -0,0 +1,38 @@
...
+        with:
+          token: ${{ steps.generate-token.outputs.token }}
+          branch: automation/sync-main-to-earlgrey-hwe
+          base: earlgrey-hwe
```
moidx, line 30: Do you have any thoughts on how `earlgrey-hwe` to `main` would work? 

Can we use `spi_device` driver developed in `earlgrey-hwe` development branch as an example?

cfrantz: I think merging back to main would be manual:
- Traits and project wide services require review at the project level.
- Earlgrey specific things (e.g. in `//target/earlgrey`) require review by earlgrey stakeholders.

.github/workflows/sync-main-to-earlgrey-hwe.yml
cfrantz: It seems like `ubuntu-latest` is common practice in this repo.  For this workflow in particular, we aren't building or testing anything in the repo where we want a compiler or set of libraries tied to the OS revision.  

I think for this workflow, ubuntu-latest is the safer choice - we're always running the workflow tooling in the latest supported OS version.

moidx: ack.

.github/workflows/sync-main-to-earlgrey-hwe.yml
moidx: ack. Sounds good. I am having a hard time thinking about how the manual process will work in practice, but I think it is something we can discuss offline. It is a good idea to have the development branch tracking main.

moidx (approved):

PR #244, opened 2026-05-30
[veer/earlgrey] Fix syscall_latency exit path

moidx:
The `syscall_latency` returned an exit code, as opposed to running in a loop. This triggered a `process_exit` call that resulted in a kernel crash. The root cause of the kernel crash is outside the scope of this change.

Fix the apps to match every other pw_kernel userspace test app: signal  completion with syscall::debug_shutdown(status), which shuts the whole system down with the test status and never enters the per-process teardown path.

- veer: re-enable measure_syscall_latency_emulator_test.
- earlgrey: add measure_syscall_latency_qemu_test (opentitan_test, qemu interface) so this exit path has CI coverage. Add debug_shutdown(status) at the end of the test to avoid crash.

Only review the last commit in this PR.

### DIFF
--- a/target/earlgrey/syscall_latency/BUILD.bazel
+++ b/target/earlgrey/syscall_latency/BUILD.bazel
@@ -8,7 +8,7 @@ load("@pigweed//pw_kernel/tooling:target_linker_script.bzl", "target_linker_scri
 load("@rules_rust//rust:defs.bzl", "rust_binary")
 load("//target/earlgrey:defs.bzl", "TARGET_COMPATIBLE_WITH")
 load("//target/earlgrey/signing/keys:defs.bzl", "FPGA_ECDSA_KEY", "SILICON_ECDSA_KEY")
-load("//target/earlgrey/tooling:opentitan_runner.bzl", "opentitan_runner")
+load("//target/earlgrey/tooling:opentitan_runner.bzl", "opentitan_runner", "opentitan_test")
 
 rust_app(
     name = "syscall_latency",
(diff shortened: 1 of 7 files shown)


target/earlgrey/syscall_latency/BUILD.bazel
```diff
@@ -116,3 +116,11 @@ opentitan_runner(
     tags = ["manual"],
     target = ":measure_syscall_latency",
 )
+
+opentitan_test(
```
cfrantz, line 120: The rules above are just `runner` rules because this isn't really a test per-se.  It measure syscall latency and prints the results to the console.  We should consider whether or not we want this to become a test and verify something about the latency.

Running this under qemu may or may not be valuable.  Qemu instruction timing is not accurate, so the latency numbers from qemu are not to be trusted.  There is probably some value in executing under qemu with respect to whether the program compiles, links and exits correctly.

third_party/caliptra/caliptra-mcu-sw/BUILD.bazel
```diff
@@ -209,6 +209,7 @@ rust_library(
     crate_root = "@caliptra_mcu_sw//:romtime/src/lib.rs",
     edition = "2021",
     rustc_flags = ["--cap-lints=allow"],
+    target_compatible_with = ["//third_party/caliptra/platforms:target_caliptra"],
```
cfrantz, line 212: This should maybe be a const in a `.bzl` file so we can say:
```
load("//third_party/caliptra/something.bzl", "TARGET_COMPATIBLE_WITH")
...
target_compatible_with = TARGET_COMPATIBLE_WITH,
```

See how its done in `//target/earlgrey/defs.bzl`.  Ok to fix in a follow-up PR.

cfrantz (approved):

PR #263, opened 2026-06-02
automation: roll pigweed every day 

cfrantz:
It appears that `dependabot` cannot roll dependencies that don't have
semantic-versioning version numbers.

The `roller.py` script looks for dependencies in MODULE.bazel with the
magic comment `# roll:<dep>`.  We query the origin of such deps and
update the git hash with the hash retrieved from the origin.

### DIFF
--- a/.github/workflows/roll-pigweed.yml
+++ b/.github/workflows/roll-pigweed.yml
@@ -1,0 +1,47 @@
+name: Roll pigweed
+
+on:
+  schedule:
+    # Runs every day at 15:00 UTC (8:00 AM PDT / 7:00 AM PST)
+    - cron: '0 15 * * *'
+  workflow_dispatch: # Allows you to trigger it manually from the Actions tab
+
+jobs:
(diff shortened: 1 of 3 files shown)


.github/workflows/roll-pigweed.yml
```diff
@@ -0,0 +1,47 @@
+name: Roll pigweed
+
+on:
+  schedule:
+    # Runs every day at 15:00 UTC (8:00 AM PDT / 7:00 AM PST)
+    - cron: '0 15 * * *'
```
moidx, line 6: Should we run the nightly tests with this workflow for coverage? OK to do in a follow up PR.

moidx (approved):
Should we revert the dependabot configuration?

.github/workflows/roll-pigweed.yml
cfrantz: Ack.  Will investigate adding nightly in a followup.

PR #278, opened 2026-06-04
Introduce the ast10x0 i3c peripheral

stevenlee7189:
# Overview

This PR aligns the AST10x0 I3C driver design with
`https://github.com/rusty1968/pac-design-patterns.git`.

# Pattern Alignment

## Confined-unsafe MMIO facade

The design follows the confined-unsafe MMIO facade pattern by concentrating the
MMIO construction contract and PAC-facing register access inside the
`I3cRegisters` layer.

## Cooperative-yield bounded-poll device

The design also follows the cooperative-yield bounded-poll pattern.

The hardware binding accepts a caller-supplied yield strategy, and polling paths
are explicitly bounded. As a result, wait behavior is portable across execution
models and timeout remains an observable, structured failure mode rather than an
unbounded hang.

# Interrupt-model exceptions

There are two intentional structural exceptions worth calling out explicitly.

## ISR-side aliasing

Interrupt handling requires an ISR-side register handle because the IRQ path
cannot directly borrow the thread-owned controller value. This alias is a narrow
exception made specifically for interrupt dispatch.

## Global dispatch and event handoff

The IRQ path also relies on a bounded per-bus global dispatch registry and
atomic event handoff. This is an interrupt-integration mechanism, not a general
ownership model for the driver.

These exceptions are deliberate and localized. They do not change the primary
architecture, which remains a runtime-selected single-driver design.

# Test Result
image


rusty1968 (changes requested):
We need to harden the error path - see PR into your branch.

target/ast10x0/peripherals/i3c/hardware.rs
```diff
@@ -0,0 +1,2135 @@
...
+            cmd.cmd_lo |= field_prep(COMMAND_PORT_SPEED, SpeedI3c::I2cFmAsI3c as u32);
+        }
+
+        let mut xfer = I3cXfer::new(&mut cmds[..]);
+        self.start_xfer(config, &mut xfer);
+
+        // On timeout `wait_xfer_complete` already recovered the engine;
+        // `xfer.ret` stays -1 and falls through to the error mapping below
+        // (same outcome as the reference's timeout path).
+        let _ = self.wait_xfer_complete(config, &mut xfer, I3C_OP_TIMEOUT_US);
```
rusty1968: # CCC transfer timeout is ignored.
Detail: In `do_ccc`, the result from `wait_xfer_complete(...)` is not checked. If the wait times out, the code later reports a generic `Invalid` error from `xfer.ret` instead of a clear timeout error.
Fix: Check the return from `wait_xfer_complete(...)`. If it returns `false`, return `Err(I3cDrvError::Timeout)` right away.

target/ast10x0/peripherals/i3c/hardware.rs
```diff
@@ -0,0 +1,2135 @@
...
+                break;
+            }
+            (self.yield_fn)(1_000);
+            left -= 1;
+        }
+
+        // Timeout: recover the engine and re-arm the IRQ sources. Recovery is
+        // best-effort — the `false` return already reports the timeout.
+        i3c_debug!(self.logger, "wait_xfer_complete: timeout");
+        let _ = self.enter_halt(true, config);
```
rusty1968: On timeout in `wait_xfer_complete`, the driver attempted recovery but did not enforce that recovery succeeded before allowing future transfers.

target/ast10x0/peripherals/i3c/hardware.rs
```diff
@@ -0,0 +1,2135 @@
...
+            .wait_for_us(I3C_OP_TIMEOUT_US, &mut self.yield_fn)
+        {
+            // Vendor C driver parity (`target_rst_worker`): an unanswered SIR
+            // means the engine may be wedged beyond a queue reset — re-run the
+            // full controller init. Side effects match the C driver: the
+            // dynamic address is dropped (the bus master must re-run DAA) and
+            // `sir_allowed_by_sw` is cleared until the next DA assignment.
+            // Best-effort; `IoError` below already reports the failure.
+            i3c_debug!(self.logger, "SIR timeout! Reset I3C controller");
+            let _ = self.init(config);
```
rusty1968: Please do not ignore the result of init here. If init fails, we should log it and return a clear recovery-failed error (or set a sticky fault) so the failure is visible to callers.

rusty1968 (changes requested):

target/ast10x0/peripherals/i3c/hardware.rs
```diff
@@ -0,0 +1,2135 @@
...
+            return Err(I3cDrvError::IoError);
+        }
+
+        if !events
```
rusty1968, line 2386: In the target_data_done timeout path, recovery is best-effort and reset_ctrl result is ignored.
If reset fails, we lose that signal and only return Timeout.
Please consider logging recovery failure explicitly and/or latching a sticky fault so later operations can fail fast instead of running on possibly bad controller state.

target/ast10x0/peripherals/i3c/hardware.rs
```diff
@@ -838,6 +872,31 @@ where
...
+            return Err(I3cDrvError::IoError);
+        }
+        Ok(())
+    }
+
+    /// Recover a timed-out transfer and re-arm its IRQs.
+    fn recover_timeout_xfer(&mut self, config: &mut I3cConfig) {
+        i3c_debug!(self.logger, "wait_xfer_complete: timeout");
+        let halt_ok = self.enter_halt(true, config).is_ok();
+        let reset_ok = self.reset_ctrl(RESET_CTRL_XFER_QUEUES).is_ok();
```
rusty1968, line 889: Something I just realized - do we require the halt before attempting reset_ok or are we issuing operations in an invalid state if the halt fails?

rusty1968 (approved):

rusty1968:
@stevenlee7189 , please resolve the conflicts when you have a chance.

target/ast10x0/peripherals/i3c/hardware.rs
stevenlee7189: Checked with the original Zephyr I3C driver author, enter_halt isn't really supposed to fail, so I kept reset_ctrl unconditional to match the vendor driver's behavior. I also added a comment in the code to explain why.

target/ast10x0/peripherals/i3c/hardware.rs
stevenlee7189: fixed

target/ast10x0/peripherals/i3c/hardware.rs
stevenlee7189: fixed

target/ast10x0/peripherals/i3c/hardware.rs
stevenlee7189: fixed

PR #280, opened 2026-06-04
Spdm application

wmaroneAMD:
This uplevels SPDM from the OCP demo branch. It does not include a hardware build target at this time. Tests within the Responder confirm VCA functionality.


services/spdm/requester/tests/crypto-client-stub/src/lib.rs
```diff
@@ -0,0 +1,92 @@
+// Licensed under the Apache-2.0 license
```
rusty1968, line 1: The crypto-client/client-stub folder should be  moved under the SPDM. This is the API Spdm poses to the ctypto providers not the api of clients that use openprot crypto server.

rusty1968 (approved):

PR #302, opened 2026-06-12
Orchestration: resiliency state machine

rusty1968:
This PR introduces the state machine component of the Orchestrator.  It is pure transition logic only, no platform I/O. 

The SM is the resiliency policy authority: it defines what happens, in what order, and under what conditions. Policy is expressed as a **Hierarchical State Machine (HSM)** that enforces correct sequencing through state transitions and entry/exit actions — not through ad-hoc logic scattered across a platform implementation.


LeeTroy:
The original ASPEED-PFR state machine has a super state concept, it helps to control BMC/Host's reset when entering or leaving T-1 (Preboot) vs T0(Runtime). It seems the current design doesn't have this in mind.

For the run action, there are three handler function for each state, i.e. enter/run/exit, and two hander function for super state, i.e. enter/exit, for example:
```mermaid
stateDiagram-v2
    [*] --> A: e0

    state Tmin1 {
        A --> B: e1
        C --> A: e4
    }

    state Tzero { 
        B --> B: e2
        B --> C: e3
    }

```
1. Event `e1` will trigger `A.exit()`, `Tmin1.exit()`, `Tzero.enter()`, `B.enter()`, `B.run()`
2. Event `e2` will trigger `B.run()`
3. Event `e3` will trigger `B.exit()`, `C.enter()`, `C.run()`
4. Event `e4` will trigger `C.exit()`, `Tzero.exit()`, `Tmin1.enter()`, `A.enter()`, `A.run()`

PR #314, opened 2026-06-20
ci: refactor self-hosted bazel setup into composite action

moidx:
Refactors the repeated setup-bazel steps in ci.yml into a local
composite action (.github/actions/setup-bazel-self-hosted).

This ensures a consistent configuration across all CI jobs, including:
- Disabling GitHub-backed caching to avoid redundancy and conflicts.
- Setting a unique output base per runner (${{ runner.workspace }}/.bazel)
  to avoid server trashing.
- Configuring local persistent cache paths in the runner's home directory.

Future updates to the self-hosted Bazel configuration can now be
managed in a single place.

### DIFF
--- a/.github/actions/setup-bazel-self-hosted/action.yml
+++ b/.github/actions/setup-bazel-self-hosted/action.yml
@@ -1,0 +1,17 @@
+# Licensed under the Apache-2.0 license
+
+name: 'Setup Bazel (Self-Hosted)'
+description: 'Configure Bazel for self-hosted runners with local persistent caching'
+runs:
+  using: "composite"
+  steps:
+    - name: Setup Bazel
+      uses: bazel-contrib/setup-bazel@0.19.0
+      with:
+        bazelisk-cache: false
+        disk-cache: false
+        repository-cache: false
+        output-base: ${{ runner.workspace }}/.bazel
+        bazelrc: |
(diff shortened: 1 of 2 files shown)


.github/actions/setup-bazel-self-hosted/action.yml
```diff
@@ -0,0 +1,17 @@
...
+    - name: Setup Bazel
+      uses: bazel-contrib/setup-bazel@0.19.0
+      with:
+        bazelisk-cache: false
+        disk-cache: false
+        repository-cache: false
+        output-base: ${{ runner.workspace }}/.bazel
+        bazelrc: |
+          build --disk_cache=/home/github-runner/.cache/bazel-disk
+          common --repository_cache=/home/github-runner/.cache/bazel-repo
```
cfrantz, line 17: If you can control which file this gets written to (e.g. `self-hosted.bazelrc`), we could use a `try-import` in the project's `.bazelrc` to import it if-and-only-if it exists.

In our `.bazelrc`, we'd add something like this:
```
# User bazelrc file
# =================
# See https://bazel.build/configure/best-practices#bazelrc-file
#
# Note: this should be at the bottom of the file, so that user-specified
# options override anything in this file
# (https://bazel.build/run/bazelrc#imports)
try-import %workspace%/self-hosted.bazelrc
```

cfrantz (approved):

PR #315, opened 2026-06-20
Add DFU firmware updates to the hwe firmware

cfrantz:
1. Implement the USB DFU protocol.
2. Integrate DFU into the HWE firmware.


target/earlgrey/util/dfu.rs
```diff
@@ -0,0 +1,361 @@
...
+                ManifestIdentifier::ROM_EXT => {
+                    util_zfmt::info!(FlashingRegion {
+                        region: "ROM_EXT",
+                        start: self.update.rom_ext_start,
+                        end: self.update.rom_ext_end,
+                    });
+                    self.flash_erase(self.update.rom_ext_start, self.update.rom_ext_end)
+                        .map_err(|_| DfuStatus::ErrErase)?;
+                    self.update.state = FwUpdateState::RomExt;
+                    self.update.next_erase = 32;
```
moidx: nit: if possible, make ROM_EXT size configurable from boot info, but ok to skip for now.

target/earlgrey/util/dfu.rs
```diff
@@ -0,0 +1,361 @@
...
+        util_zfmt::info!(DfuManifest);
+        if self.update.state == FwUpdateState::Done {
+            // TODO: check for errors.
+            let _ = self
+                .sysmgr
+                .set_boot_policy(earlgrey_sysmgr_client::BootPolicy {
+                    pref_slot: self.update.app,
+                    next_slot: BootSlot::Unspecified,
+                });
+            let _ = self.sysmgr.request_reboot();
```
moidx, line 355: from AI review: `bmAttributes` set `bitManifestationTolerant`. Should we consider setting `bitWillDetach` as well given that this function requests DUT reboot?

target/earlgrey/util/dfu.rs
```diff
@@ -0,0 +1,361 @@
...
+                    self.update.start_block = block_num;
+                }
+                _ => {
+                    util_zfmt::error!(UnknownManifestId { id: id.0 as u32 });
+                    return Err(DfuStatus::ErrUnknown);
+                }
+            }
+        }
+
+        let block = block_num - self.update.start_block;
```
moidx, line 275: do we need any data size bound checks against flash partition target size?

target/earlgrey/util/dfu.rs
```diff
@@ -0,0 +1,361 @@
...
+    let (partition, mut n) = match n {
+        0 => (0, 0), // The UDS (dice) cert is located in bank=0, page=9.
+        1 => (1, 0), // The CDI (dice) certs are located in bank=1, page=9.
+        2 => (1, 1), // CDI1 is the second cert in bank=1, page=9.
+        _ => return Err(DfuStatus::ErrFile),
+    };
+    let mut offset = 0usize;
+    let mut buf = [0u8; 1024];
+    loop {
+        let sz = core::cmp::min(2048 - offset, buf.len());
```
moidx, line 91: Is it worth checking that `offset` does not cause an underflow, or is this safe based on the type of `sz`. I suspect that this could only fail if the data is somehow corrupted, but that will result in a different failure at boot time?

moidx (approved):

target/earlgrey/util/dfu.rs
cfrantz: Added bounds checks.

target/earlgrey/util/dfu.rs
cfrantz: We already set `bitManifestationTolerant` in `protocol/usb/dfu/lib.rs`.

I don't think `bitWillDetach` is appropriate for us because we aren't transitioning from App mode to DFU mode with a `DFU_DETACH` command.

target/earlgrey/util/dfu.rs
cfrantz: Fixed.

target/earlgrey/util/dfu.rs
cfrantz: This is safe because `offset` is unsigned; an underflow will cause `min` to choose `buf.len()` as the size.

PR #332, opened 2026-07-01
earlgrey: create base for transport firmware 

cfrantz:


target/earlgrey/firmware/transport/target.rs
```diff
@@ -0,0 +1,33 @@
...
+
+#![no_std]
+#![no_main]
+use target_common::{declare_target, TargetInterface};
+use {console_backend as _, entry as _};
+
+pub struct Target {}
+
+impl TargetInterface for Target {
+    const NAME: &'static str = "Earlgrey Hardware Enablement Firmware";
```
moidx: Should we call this one `Earlgrey Transport Firmware`?

target/earlgrey/firmware/transport/logmgr.rs
```diff
@@ -0,0 +1,263 @@
...
+}
+
+fn logmgr_server() -> Result<(), Error> {
+    // UART0 physical address is mapped in our address space.
+    // Since we use identity mapping for devices, we can use the physical address directly.
+    let mut uart = unsafe { UartDriver::new(Uart0::PTR) };
+
+    // Configure UART: 115200 baud, 8N1.
+    uart.configure(UsartConfig {
+        baud_rate: 0,
```
moidx, line 164: Is this correct?

moidx (approved):

target/earlgrey/firmware/transport/logmgr.rs
cfrantz: Yes.  The uart driver is setup to accept a baud rate of `0` to mean "don't change the existing NCO value", which means leave the baud rate configured as the ROM_EXT configured it.

target/earlgrey/firmware/transport/target.rs
cfrantz: Good catch.  Fixed.

PR #334, opened 2026-07-06
earlgrey: Implement SPI driver and SPI Flash Server

lchiawei:
This is for early review.

These commits are modified from the internal implementation. The main change I made is mentioned in the message for each commit.

### DIFF
--- a/drivers/flash/spi_flash.rs
+++ b/drivers/flash/spi_flash.rs
@@ -1,0 +1,1982 @@
+// Licensed under the Apache-2.0 license
+// SPDX-License-Identifier: Apache-2.0
+
+#![no_std]
+
+use bitfield_struct::bitfield;
+use core::cmp::min;
+use core::convert::TryFrom;
+use core::num::NonZero;
+use core::prelude::v1::*;
+use util_sfdp::{QuadEnableRequirements, SfdpReader};
+use util_types::PowerOf2Usize;
+use zerocopy::{FromBytes, FromZeros, Immutable, IntoBytes};
+
+use embedded_hal::spi::Operation;
+use hal_flash::{Flash as FlashTrait, FlashAddress};
+use util_error::{self as error, ErrorCode};
+use util_io::RandomRead;
+
+// TODO(b/481400917): Replace with stronger "byte count" type.
+const KIB: usize = 1024;
+const MIB: usize = 1024 * KIB;
+
+// The maximum number of bytes the opcode + address + dummy bytes at the start
+// of a transaction will need.
+const MAX_PREFIX_LEN: usize = 6;
+const MAX_3B_SIZE: usize = 16 * MIB;
+const SECTOR_SIZE: usize = 4096;
+const BLOCK_SIZE: usize = 64 * KIB;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SpiTxnWidth {
+    STANDARD = 0,
+    DUAL = 1,
+    QUAD = 2,
+}
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct SfCmd {
+    pub opcode: u8,
+    pub width: SpiTxnWidth,
+    pub addr_mode: AddressingMode,
+}
+
+impl SfCmd {
+    pub const READ: Self = Self {
+        opcode: OP_READ,
+        width: SpiTxnWidth::STANDARD,
+        addr_mode: AddressingMode::_3Byte,
+    };
+    pub const PROGRAM: Self = Self {
+        opcode: OP_PROGRAM,
+        width: SpiTxnWidth::STANDARD,
+        addr_mode: AddressingMode::_3Byte,
+    };
+    pub const ERASE: Self = Self {
+        opcode: OP_ERASE_4K,
+        width: SpiTxnWidth::STANDARD,
+        addr_mode: AddressingMode::_3Byte,
+    };
+    pub const READ4B: Self = Self {
+        opcode: OP_READ4B,
+        width: SpiTxnWidth::STANDARD,
+        addr_mode: AddressingMode::_4Byte,
+    };
+    pub const PROGRAM4B: Self = Self {
+        opcode: OP_PROGRAM4B,
+        width: SpiTxnWidth::STANDARD,
+        addr_mode: AddressingMode::_4Byte,
+    };
+    pub const ERASE4B: Self = Self {
+        opcode: OP_ERASE4B_4K,
+        width: SpiTxnWidth::STANDARD,
+        addr_mode: AddressingMode::_4Byte,
+    };
+    pub const ERASE64K: Self = Self {
+        opcode: OP_ERASE_64K,
+        width: SpiTxnWidth::STANDARD,
+        addr_mode: AddressingMode::_3Byte,
+    };
+    pub const ERASE4B_64K: Self = Self {
+        opcode: OP_ERASE4B_64K,
+        width: SpiTxnWidth::STANDARD,
+        addr_mode: AddressingMode::_4Byte,
+    };
+    pub const QREAD: Self = Self {
+        opcode: OP_QREAD,
+        width: SpiTxnWidth::QUAD,
+        addr_mode: AddressingMode::_3ByteWithDummy,
+    };
+    pub const QREAD4B: Self = Self {
+        opcode: OP_QREAD4B,
+        width: SpiTxnWidth::QUAD,
+        addr_mode: AddressingMode::_4ByteWithDummy,
+    };
+    pub const QPROGRAM: Self = Self {
+        opcode: OP_QPROGRAM,
+        width: SpiTxnWidth::QUAD,
+        addr_mode: AddressingMode::_3Byte,
+    };
+    pub const QPROGRAM4B: Self = Self {
+        opcode: OP_QPROGRAM4B,
+        width: SpiTxnWidth::QUAD,
+        addr_mode: AddressingMode::_4Byte,
+    };
+    pub const READ4B_GLOBAL: Self = Self {
+        opcode: OP_READ,
+        width: SpiTxnWidth::STANDARD,
+        addr_mode: AddressingMode::_4Byte,
+    };
+    pub const PROGRAM4B_GLOBAL: Self = Self {
+        opcode: OP_PROGRAM,
+        width: SpiTxnWidth::STANDARD,
+        addr_mode: AddressingMode::_4Byte,
+    };
+    pub const ERASE4B_GLOBAL: Self = Self {
+        opcode: OP_ERASE_4K,
+        width: SpiTxnWidth::STANDARD,
+        addr_mode: AddressingMode::_4Byte,
+    };
+    pub const ERASE64K_GLOBAL: Self = Self {
+        opcode: OP_ERASE_64K,
+        width: SpiTxnWidth::STANDARD,
+        addr_mode: AddressingMode::_4Byte,
+    };
+}
+
+pub struct SpiFlashConfig {
+    pub read: SfCmd,
+    pub program: SfCmd,
+    pub erase4k: SfCmd,
+    pub erase64k: SfCmd,
+
+    /// The total size of the flash in bytes
+    pub size: NonZero<usize>,
+    pub quad_enable_req: Option<QuadEnableRequirements>,
+}
+
+impl SpiFlashConfig {
+    pub fn from_sfdp_conservative<R: RandomRead<Error = ErrorCode>>(
+        sfdp_bytes: R,
+    ) -> Result<Self, ErrorCode> {
+        let mut sfdp = SfdpReader::new(sfdp_bytes)?;
+        let table = sfdp.basic_flash_parameters()?;
+
+        let size = usize::try_from(table.table_jesd216().memory_density.byte_len()?).unwrap();
+        let size = NonZero::new(size).ok_or(error::FLASH_GENERIC_INVALID_SIZE)?;
+        let config = if size.get() <= MAX_3B_SIZE {
+            SpiFlashConfig {
+                size,
+                read: SfCmd::READ,
+                erase4k: SfCmd::ERASE,
+                erase64k: SfCmd::ERASE64K,
+                program: SfCmd::PROGRAM,
+                quad_enable_req: None,
+            }
+        } else {
+            SpiFlashConfig {
+                size,
+                read: SfCmd::READ4B_GLOBAL,
+                erase4k: SfCmd::ERASE4B_GLOBAL,
+                erase64k: SfCmd::ERASE64K_GLOBAL,
+                program: SfCmd::PROGRAM4B_GLOBAL,
+                quad_enable_req: None,
+            }
+        };
+        Ok(config)
+    }
+
+    pub fn from_sfdp<R: RandomRead<Error = ErrorCode>>(sfdp_bytes: R) -> Result<Self, ErrorCode> {
+        let mut sfdp = SfdpReader::new(sfdp_bytes)?;
+        let table = sfdp.basic_flash_parameters()?;
+
+        let size = usize::try_from(table.table_jesd216().memory_density.byte_len()?).unwrap();
+        let size = NonZero::new(size).ok_or(error::FLASH_GENERIC_INVALID_SIZE)?;
+
+        // TODO: Restore Quad SPI (QSPI) support once earlgrey_spi_host implements
+        // DUAL/QUAD transaction widths.
+        let config = if size.get() <= MAX_3B_SIZE {
+            SpiFlashConfig {
+                size,
+                read: SfCmd::READ,
+                erase4k: SfCmd::ERASE,
+                erase64k: SfCmd::ERASE64K,
+                program: SfCmd::PROGRAM,
+                quad_enable_req: None,
+            }
+        } else {
+            SpiFlashConfig {
+                size,
+                read: SfCmd::READ4B_GLOBAL,
+                erase4k: SfCmd::ERASE4B_GLOBAL,
+                erase64k: SfCmd::ERASE64K_GLOBAL,
+                program: SfCmd::PROGRAM4B_GLOBAL,
+                quad_enable_req: None,
+            }
+        };
+        Ok(config)
+    }
+}
+
+/// "Driver" for SPI NOR flash.
+pub struct SpiFlash<S: embedded_hal::spi::SpiDevice> {
+    spi: S,
+    config: SpiFlashConfig,
+    initialized: bool,
+}
+
+impl<S: embedded_hal::spi::SpiDevice> SpiFlash<S> {
+    pub fn new(spi: S) -> Self {
+        Self {
+            spi,
+            config: SpiFlashConfig {
+                read: SfCmd::READ,
+                program: SfCmd::PROGRAM,
+                erase4k: SfCmd::ERASE,
+                erase64k: SfCmd::ERASE64K,
+                size: NonZero::new(1).unwrap(),
+                quad_enable_req: None,
+            },
+            initialized: false,
+        }
+    }
+
+    pub fn init(&mut self) -> Result<(), ErrorCode> {
+        let sfdp_bytes = SfdpRandRead { spi: &mut self.spi };
+        let config = SpiFlashConfig::from_sfdp(sfdp_bytes)?;
+        let qer = config.quad_enable_req;
+
+        let mut status = Status::new_zeroed();
+        self.transfer_req_resp(&[OP_STATUS], status.as_mut_bytes())?;
+
+        if status.busy() {
+            self.wait_for_busy_to_clear()?;
+        }
+
+        if let Some(qer) = qer {
+            match qer {
+                QuadEnableRequirements::NoQeBit => {}
+                QuadEnableRequirements::QeBit6SR1 => {
+                    if !status.maybe_quad_en() {
+                        status.set_maybe_quad_en(true);
+                        self.transfer_req_resp(&[OP_WRITE_EN], &mut [])?;
+                        self.transfer_req_resp(&[OP_WR_STATUS, status.into()], &mut [])?;
+                    }
+                }
+                _ => {
+                    self.config.read = SfCmd::READ4B;
+                }
+            }
+        }
+
+        if config.size.get() > 16 * MIB {
+            self.enter_4byte_mode()?;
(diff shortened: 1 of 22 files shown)


util/sfdp/mod.rs
```diff
@@ -0,0 +1,2297 @@
+// Licensed under the Apache-2.0 license
+// SPDX-License-Identifier: Apache-2.0
+
+#![cfg_attr(not(test), no_std)]
+
+const _: () = assert!(cfg!(target_endian = "little"));
+
+use bitfield_struct::bitfield;
+use core::mem::offset_of;
+use core::time::Duration;
```
cfrantz: We should investigate `pw_time::Duration` and see if its a better fit.

target/earlgrey/drivers/spi_host.rs
```diff
@@ -0,0 +1,296 @@
...
+const MAX_RX_CHUNK_LEN: usize = 256;
+const MAX_TX_CHUNK_LEN: usize = 288;
+const DIR_RXONLY: u32 = 1;
+const DIR_TXONLY: u32 = 2;
+
+impl<Mmio: SpiHostMmio> embedded_hal::spi::ErrorType for SpiHost<Mmio> {
+    type Error = SpiError;
+}
+
+impl<Mmio: SpiHostMmio> embedded_hal::spi::SpiDevice for SpiHost<Mmio> {
```
cfrantz: Does embedded_hal's SPI traits take into account different spi transaction widths? (Single, Dual, Quad)?

drivers/flash/spi_flash.rs
```diff
@@ -0,0 +1,396 @@
...
+            .map_err(|_| error::FLASH_GENERIC_SFDP_INVALID_MEMORY_DENSITY)?;
+
+        self.size_bytes = size as usize;
+        self.erasable_sizes_bitmap = (1 << 12) | (1 << 16); // Support 4KB (bit 12) and 64KB (bit 16)
+
+        // Verify page size if JESD216A+ is supported
+        if let Some(jesd216a) = table.table_jesd216a() {
+            let page_size = jesd216a.word11.page_size().value();
+            if page_size != 256 {
+                pw_log::error!(
```
cfrantz: We shouldn't log in the spi_flash driver.

drivers/flash/spi_flash.rs
```diff
@@ -0,0 +1,396 @@
...
+    }
+
+    fn format_cmd_prefix(opcode: u8, addr: u32, addr_bytes: u8) -> ([u8; 5], usize) {
+        let [b0, b1, b2, b3] = addr.to_be_bytes();
+        if addr_bytes == 3 {
+            ([opcode, b1, b2, b3, 0], 4)
+        } else {
+            ([opcode, b0, b1, b2, b3], 5)
+        }
+    }
```
cfrantz: Returning the fixed sized array and size like this requires a checked slicing operation.
We can avoid the checked slice by returning a slice from the command formatting function:
```
fn format_cmd_prefix(buf: &mut [u8; MAX_PREFIX], opcode: u8, addr: u32, addr_size: u8) -> &[u8] {
    let [b0, b1, b2, b3] = addr.to_be_bytes();
    buf[0] = opcode;
    if addr_bytes == 3 {
        buf[1] = b1; buf[2] = b2; buf[3] = b3;
        &buf[..4]
    } else {
        buf[1] = b0; buf[2] = b1; buf[3] = b2; buf[4] = b3;
        &buf[..5]
    }
}
```

drivers/flash/spi_flash.rs
```diff
@@ -0,0 +1,396 @@
...
+        );
+
+        for _ in 0..self.read_cmd.dummy_bytes {
+            if let Some(cell) = prefix.get_mut(len) {
+                *cell = 0;
+                len = len
+                    .checked_add(1)
+                    .ok_or(error::FLASH_GENERIC_BAD_ALIGNMENT)?;
+            }
+        }
```
cfrantz: The command formatting could be made into a function on `SfCmd` and we could avoid repeating this sequence of code for the various operations.

target/earlgrey/drivers/spi_host.rs
```diff
@@ -0,0 +1,296 @@
...
+    }
+    fn regs_mut(&mut self) -> spi_host::RegisterBlock<ureg::RealMmioMut<'_>> {
+        self.regs_mut()
+    }
+}
+
+/// Earlgrey SPI Host driver.
+pub struct SpiHost<Mmio: SpiHostMmio> {
+    mmio: Mmio,
+}
```
cfrantz: We should try to note use these extra traits.  Instead, we should declare `SpiHost` something like this:

```
type SpiHostMmio = spi_host::RegisterBlock<ureg::RealMmioMut>;

pub struct SpiHost<M: SpiHostMmio> {
    mmio: M,
}
```

We need to do a bit of work to understand if we need any lifetime annotations.  I think `'static` is appropriate as it is going to exclusively own the peripheral.

target/earlgrey/drivers/spi_host.rs
cfrantz: We can stick with embedded_hal for now.  In the future, we'll consider extending the hal or writing own extension traits to handle Dual and Quad modes.

lchiawei:
Also rolled back the unrelated `services/flash/server.rs` to make this PR more focused.

target/earlgrey/drivers/spi_host.rs
lchiawei: Yeah, I believe that `embedded_hal` only has single mode. Do we have requirements for DUAL and QUAD modes in foreseeable milestones?

drivers/flash/spi_flash.rs
lchiawei: Removed all the occurrences using `pw_log` in this file.

drivers/flash/spi_flash.rs
lchiawei: Done. Addressed it together with the comment below.

drivers/flash/spi_flash.rs
lchiawei: Done. Also prefilled the buffer with 0s so that we do not have to fill the dummy bytes anymore.

util/sfdp/mod.rs
lchiawei: Can we just stick to use `core::time::Duration` in this PR because we don't really utilize the timeout in our driver? Also, one of the reason that I use `core` library here is that I would like to make the library and unit tests in this file to be less platform-dependent and easily tested on host. But I agreed that we should care about whether it costs too much to perform (microseconds * clock) on the hardware.

util/sfdp/mod.rs
```diff
@@ -0,0 +1,2297 @@
...
+            .checked_mul(size_of::<u32>())
+            .ok_or(error::FLASH_GENERIC_SFDP_PARAMETERS_TOO_LONG)?
+            .min(target_slice.len());
+        if bytes_len < min_bytes {
+            return Err(error::FLASH_GENERIC_SFDP_PARAMETERS_TOO_SHORT.into());
+        }
+
+        let target = target_slice
+            .get_mut(..bytes_len)
+            .ok_or(error::FLASH_GENERIC_SFDP_PARAMETERS_TOO_LONG)?;
```
cfrantz: Was this flagged by the panic detector?  The `min` expression above should have already proved to the optimizer that `bytes_len` can never be larger than the size of the target_slice.

util/sfdp/mod.rs
cfrantz: `pw_time` is a single 64-bit integer, has a host implementation and it allows you to choose an arbitrary tick rate by implementing `Clock` on a type.

`core::time::Duration` has a more complex representation:
```
pub struct Duration {
    secs: u64,
    nanos: Nanoseconds, // Always 0 <= nanos < NANOS_PER_SEC
}
```
Math ops on this type will always involve the 64-bit computation and whatever `Nanoseconds` is (probably a u32) and will involve managing carry/borrow between the two components.

I think for now, we should do what the original code did: create a `Nanoseconds` type alias over a nanosecond resolution clock:

```
pub struct NanoClock { ... }
impl pw_time::Clock for NanoClock {
    const TICKS_PER_SEC: u64 = 1_000_000_000;
}

type Nanoseconds = pw_time::Duration<NanoClock>;
```

This strategy will keep the complexity of the `core` type out of this code but still allow precise extraction of the time value if needed.

target/earlgrey/drivers/spi_host.rs
cfrantz: My previous comment may have gotten lost.  These traits (which exist in the original) aren't really needed.  We need only to keep a `spi_host::RegisterBlock<ureg::RealMmioMut<'_>>` in `struct SpiHost`.  We can get or construct that entity from the spi_host registers crate.

target/earlgrey/drivers/spi_host.rs
```diff
@@ -0,0 +1,296 @@
...
+            }
+
+            let mut txqd_timeout = TIMEOUT_LIMIT;
+            while regs.status().read().txqd() != 0 {
+                txqd_timeout = txqd_timeout.checked_sub(1).ok_or(SpiError::Timeout)?;
+            }
+
+            let chunk_len = core::cmp::min(data.len(), MAX_TX_CHUNK_LEN);
+            let chunk = data.get(..chunk_len).ok_or(SpiError::InvalidTransaction)?;
+            data = data.get(chunk_len..).ok_or(SpiError::InvalidTransaction)?;
```
cfrantz: It seems like we're performing the same slicing operation twice here.  It also seems that `min` should allow us to do the slicing operation without needing the `get`.

Is this something flagged by the no_panics_test?

drivers/flash/spi_flash.rs
```diff
@@ -0,0 +1,400 @@
...
+
+#[derive(Clone, Copy)]
+struct SfCmd {
+    opcode: u8,
+    addr_bytes: u8,
+    dummy_bytes: u8,
+}
+
+impl SfCmd {
+    fn format_prefix<'a>(&self, buf: &'a mut [u8], addr: u32) -> Result<&'a [u8], ErrorCode> {
```
cfrantz: If you always pass an exact sized `buf`, you are passing an array instead of a slice.  Since arrays have known sizes, there are certain panic checks that can proven at compile time.  I recommend establishing a max prefix size (e.g. 6 bytes), making that a constant and then using buffers of that size everywhere they are needed.

drivers/flash/spi_flash.rs
```diff
@@ -0,0 +1,400 @@
...
+        let len = 1usize
+            .checked_add(self.addr_bytes as usize)
+            .and_then(|l| l.checked_add(self.dummy_bytes as usize))
+            .ok_or(error::FLASH_GENERIC_BAD_ALIGNMENT)?;
+        let prefix = buf
+            .get_mut(..len)
+            .ok_or(error::FLASH_GENERIC_BAD_ALIGNMENT)?;
+
+        // Fill prefix with 0 to initialize dummy bytes segment automatically.
+        prefix.fill(0);
```
cfrantz: If the prefix buffer is always declared in the caller like `let mut buffer = [0u8; PREFIX_LEN]`, you don't need to fill it with zeros.  You can make it part of the contract for this function that the caller needs to pre-initialize the buffer to zero.

drivers/flash/spi_flash.rs
```diff
@@ -0,0 +1,400 @@
...
+        *first = self.opcode;
+
+        match self.addr_bytes {
+            3 => {
+                let addr_slice: &mut [u8; 3] = rest
+                    .get_mut(..3)
+                    .ok_or(error::FLASH_GENERIC_BAD_ALIGNMENT)?
+                    .try_into()
+                    .map_err(|_| error::FLASH_GENERIC_BAD_ALIGNMENT)?;
+                *addr_slice = [b1, b2, b3];
```
cfrantz: This also seems like a lot of error checking for what is poking 3 or 4 bytes into a buffer that should be of a known and correct size.

drivers/flash/spi_flash.rs
```diff
@@ -0,0 +1,400 @@
...
+    fn write_enable(&mut self) -> Result<(), ErrorCode> {
+        self.spi
+            .write(&[OP_WRITE_EN])
+            .map_err(|_| error::FLASH_GENERIC_BUSY)?;
+        Ok(())
+    }
+
+    pub fn read_status(&mut self) -> Result<u8, ErrorCode> {
+        let tx = [OP_READ_STATUS];
+        let mut rx = [0u8];
```
cfrantz: A neat trick you can do instead of dealing with 1-byte arrays:
```
let mut rx = 0u8;
let mut ops = [
  embedded_hal::spi::Operation::Read(core::slice::from_mut(&mut rx)),
];
```

This works because the "slice" type can be thought of as a built-in type in rust and it can construct a slice-of-1 from a ref-to-item.

target/earlgrey/drivers/spi_host.rs
lchiawei: I found that SPI Host 0 and 1 have literally identical type `spi_host::RegisterBlock<ureg::RealMmioMut>` instead of just having a same trait, so we can even avoid the generics.

lchiawei:
I found myself missing the boundary check for `program()` (as what I did in `erase()/read()`) in `spi_flash`, so I just added it.
Besides, I also added a QEMU version for SPI host smoke test to make it covered by our presubmit CI.

drivers/flash/spi_flash.rs
```diff
@@ -0,0 +1,400 @@
...
+        let mut addr = start_addr;
+
+        while !data.is_empty() {
+            let chunk_len = core::cmp::min(
+                data.len(),
+                PROGRAM_WINDOW_SIZE - ((addr.offset() & window_mask as u32) as usize),
+            );
+            let chunk = data
+                .get(..chunk_len)
+                .ok_or(error::FLASH_GENERIC_BAD_ALIGNMENT)?;
```
cfrantz: This should also be provable because of the `min` operation.

target/earlgrey/drivers/spi_host.rs
cfrantz: I misread the two `get` operations last night.  The `min` operation should prove to the optimizer that `chunk_len` is in bounds, and we can use `split` to get the two slices we want:
```
            let (chunk, rest) = data.split_at(chunk_len);
            data = rest;
```

target/earlgrey/drivers/spi_host.rs
```diff
@@ -0,0 +1,273 @@
...
+    }
+
+    /// Initialize the SPI Host peripheral.
+    ///
+    /// Configures the clock, SPI mode (0), CS timings, performs a reset,
+    /// and enables the peripheral.
+    pub fn init(&mut self) -> Result<(), SpiError> {
+        // Default safe configuration (24MHz if core is 96MHz, Mode 0, CS0)
+        let default_config = SpiConfig {
+            clkdiv: 1,
```
cfrantz: spi_host_0 and spi_host_1 are in different clock domains:
- spi_host_0 is in the high-speed peripheral domain
- spi_host_1 is in the normal peripheral domain

We need to pass different dividers to initialize them to the same speed.

target/earlgrey/drivers/spi_host.rs
```diff
@@ -0,0 +1,273 @@
...
+            cpha: false,
+            csid: 0,
+        };
+        self.configure(&default_config)?;
+
+        // Reset the peripheral.
+        self.mmio.control().write(|w| w.sw_rst(true));
+
+        // Wait for both FIFOs to empty.
+        let mut timeout = TIMEOUT_LIMIT;
```
cfrantz: I don't think we need these timeout loops: a timeout in these cases mean the hardware is broken and isn't clocking data through the peripheral.

target/earlgrey/drivers/spi_host.rs
```diff
@@ -0,0 +1,273 @@
...
+
+        while !data.is_empty() {
+            let mut timeout = TIMEOUT_LIMIT;
+            loop {
+                let status = self.mmio.status().read();
+                if status.ready() && !status.active() {
+                    break;
+                }
+                timeout = timeout.checked_sub(1).ok_or(SpiError::Timeout)?;
+            }
```
cfrantz: Similarly, a timeout here means bad hardware.  This is the place a block-for-interrupt belongs.  We can busy-loop for now and add interrupt support in the future (leave a todo):
```
while !self.mmio.regs_mut().status().read().ready() {
    // TODO: we should block here for and interrupt that would signal the hardware is ready.
}
```

target/earlgrey/drivers/spi_host.rs
```diff
@@ -0,0 +1,273 @@
...
+
+        while len > 0 {
+            let mut timeout = TIMEOUT_LIMIT;
+            loop {
+                let status = self.mmio.status().read();
+                if status.ready() && !status.active() {
+                    break;
+                }
+                timeout = timeout.checked_sub(1).ok_or(SpiError::Timeout)?;
+            }
```
cfrantz: Busy-loop for now.

target/earlgrey/drivers/spi_host.rs
```diff
@@ -0,0 +1,273 @@
...
+
+            // Prevent out-of-bounds panics on split_at_mut.
+            if chunk_len > dest.len() {
+                return Err(SpiError::InvalidTransaction);
+            }
+            // Split dest into the chunk we want to read now, and the rest for later.
+            let (mut chunk_dest, rest) = dest.split_at_mut(chunk_len);
+            dest = rest;
+
+            let mut rx_timeout = TIMEOUT_LIMIT;
```
cfrantz: Busy-loop for now.

target/earlgrey/drivers/spi_host.rs
```diff
@@ -0,0 +1,273 @@
...
+            while !chunk_dest.is_empty() {
+                let status = self.mmio.status().read();
+                let rxqd = status.rxqd() as usize; // rxqd is in 32-bit words
+                let bytes_in_fifo = rxqd.checked_mul(4).ok_or(SpiError::HardwareError)?;
```
cfrantz: Is the checked_mul necessary?  We know from the hardware spec that the fifo depth can have a certain maximum size (8 bits in the `status` register).  We can never get a value from the hardware that would overflow.   We also `min` this value against the `chunk_dest.len()`, so we're bounding the length appropriately.

target/earlgrey/tests/drivers/spi_host/spi_host.rs
cfrantz, line 1: nit: we should put this test in `//target/earlgrey/tests/drivers/spi_host/`

target/earlgrey/tests/spi_flash_smoke/spi_flash_smoke.rs
```diff
@@ -0,0 +1,125 @@
...
+
+// Polling-based Blocking implementation for the test.
+struct SpiFlashSleep;
+
+impl Blocking for SpiFlashSleep {
+    fn wait_for_notification(&self) {
+        // Sleep for 1 millisecond between WIP status checks to reduce CPU load
+        let _ = sleep_until(SystemClock::now() + Duration::from_millis(1));
+    }
+}
```
cfrantz: Sleeping while waiting for the SPI flash is more properly the responsibility of the `spi_host` driver and the sleep should be interrupt driven.

drivers/flash/spi_flash.rs
cfrantz, line 1: Lets do a comparison of this driver vs. the original mutask driver.

lchiawei: Let me reorganize the `spi_flash.rs` here so that we can perform an easier line-to-line comparison with the original mutask implementation

target/earlgrey/tests/drivers/spi_host/spi_host.rs
lchiawei: Done.

target/earlgrey/drivers/spi_host.rs
lchiawei: Done.

target/earlgrey/drivers/spi_host.rs
lchiawei: Done. To make user's life simpler, I also add two const SpiConfig `DEFAULT_SPI0` & `DEFAULT_SPI1`. Original implementation hard codes different `clkdiv` values when initializing SpiHost0/SpiHost1, but I think we prefer to extract them and make them configurable here?

target/earlgrey/drivers/spi_host.rs
lchiawei: Done.

target/earlgrey/drivers/spi_host.rs
lchiawei: I come up with a question: If we accept `spi_host::RegisterBlock<ureg::RealMmioMut<'_>>` or similar things directly instead of a token for constructor, how can the compiler make sure the user doesn't initialize it twice?

util/sfdp/mod.rs
lchiawei: Done. However, since neither `pw_time::Duration` nor `core::ops::Mul` are owned by ourselves, we cannot overload the `*` operator and need to implement our own `mul` method.

target/earlgrey/drivers/spi_host.rs
lchiawei: Done. We can simply use `*` here.

target/earlgrey/drivers/spi_host.rs
lchiawei: Done.

target/earlgrey/drivers/spi_host.rs
lchiawei: Done.

target/earlgrey/drivers/spi_host.rs
lchiawei: Done.

util/sfdp/mod.rs
lchiawei: Done. We can use the simple slicing here.

drivers/flash/spi_flash.rs
lchiawei: Hi @cfrantz, I just ported the entire SPI flash from the original project to here, and we should be able to review it by simply diffing these two files. (Unit tests are still WIP.)

Also, some of the comments above may become obsolete due to this re-implementation.

lchiawei:
Close this PR for now as I split and moved it to #345, #346 and #347.

PR #335, opened 2026-07-08
fix(ast10x0-i2c): handle buffer-mode rx and smbus probe

pojuiatwork3881154:
Fix AST1060 I2C slave-mode status handling that could hang the bus.

- Handle buffer-mode RX_DONE (no WAIT_RX_DMA): report DataReceived.
- Handle bare SLAVE_MATCH (SMBus quick probe): re-arm and clear status so the slave does not clock-stretch until SMBus timeout.


JesseMelon (approved):

JesseMelon:
Thank you for your contribution. If you mark this as ready for review it will be merged.

PR #342, opened 2026-07-14
earlgrey: add the `Manifest` structure

cfrantz:


target/earlgrey/util/manifest.rs
```diff
@@ -0,0 +1,477 @@
...
+        if DIGEST_REGION_START_OFFSET <= end && end <= image_bytes.len() {
+            Some(&image_bytes[DIGEST_REGION_START_OFFSET..end])
+        } else {
+            None
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
moidx, line 391: Can you add a test target so that this runs in pre-submit?

moidx (approved):

target/earlgrey/util/manifest.rs
```diff
@@ -0,0 +1,477 @@
...
+    /// digest computation.
+    ///
+    /// The digest region starts immediately after the `usage_constraints` field
+    /// and ends at `signed_region_end`.
+    ///
+    /// Returns `Some(&[u8])` if the bounds are valid and within `image_bytes`,
+    /// otherwise `None`.
+    pub fn digest_region<'a>(&self, image_bytes: &'a [u8]) -> Option<&'a [u8]> {
+        // Offset of usage_constraints (384) + size of usage_constraints (48) = 432.
+        const DIGEST_REGION_START_OFFSET: usize = 432;
```
anthonychen1251: Would it be better to use compile-time offset calculation instead of hardcoding 432?
e.g.
```
        let digest_region_start_offset = offset_of!(Self, ecdsa_public_key);
```

target/earlgrey/util/manifest.rs
cfrantz: Done.

target/earlgrey/util/manifest.rs
cfrantz: I'm going to eliminate this and the other helper functions.  They're written as though we're executing in the ROM_EXT and have access to all of flash as memory.  This is not the usage model in our firmware.

anthonychen1251 (approved):

PR #343, opened 2026-07-14
earlgrey: update signing infrastructure

cfrantz:
Update signing tooling to incorporate ownershp transfer blob manifest extensions.


cfrantz:
Note: this PR requires an updated devbundle to be functional.

moidx (approved):

PR #345, opened 2026-07-14
util/sfdp: Port SFDP parser from legacy implementation

lchiawei:
Port the Serial Flash Discoverable Parameters (SFDP) parser library from the legacy implementation.

Key modifications and fixes made during the port:
- Time representation: Replaced the custom `Nanoseconds` type with new custom type based on `core::time::Duration` across all timing tables.
- Legacy Bug Fixes: Corrected incorrect bitmasks in `ChipEraseTimeUnit::from_bits` and `DurationEnumAUnit::from_bits` from `& 0b1` to `& 0b11`, enabling the parser to correctly resolve all 4 states instead of triggering panics.
- Default Value in Enums: Eliminated all `unreachable!()` wildcard arms in enum deserializers, replacing them with catch-all wildcards returning default or safe fallback states.
- Underflow & Overflow Protection:
  - Added capacity bounds checks in `MemoryDensity::byte_len` to prevent subtraction underflow when bit length is less than 3.
  - Added shift boundary checks in `PowerOf2::value` to prevent undefined behavior/panics when exponent exceeds platform `usize` bit width.
- Code Cleanups: Refactored `U24` to use `to_le_bytes`/`from_le_bytes` instead of raw bitwise shifts, and streamlined `MemoryDensity::from_byte_len` with pattern matching.

Also includes original unit tests running against mock SFDP tables.

### DIFF
--- a/util/types/time.rs
+++ b/util/types/time.rs
@@ -1,0 +1,29 @@
+// Licensed under the Apache-2.0 license
+// SPDX-License-Identifier: Apache-2.0
+
+#[derive(Debug)]
+pub struct NanoClock;
+
+impl pw_time::Clock for NanoClock {
+    const TICKS_PER_SEC: u64 = 1_000_000_000;
+
+    // NanoClock is only used as a compile-time tick configuration for Duration conversion.
+    // now() is implemented with a dummy value only to satisfy the Clock trait requirement.
+    fn now() -> pw_time::Instant<Self> {
+        pw_time::Instant::from_ticks(0)
+    }
+}
+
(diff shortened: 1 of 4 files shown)


util/types/time.rs
cfrantz, line 1: FYI: we should upstream a change to pigweed to add multiply to their `Duration` type.  After we get that upstream, we can eliminate this implementation.

cfrantz (approved):

util/types/time.rs
lchiawei: Acked.

PR #346, opened 2026-07-14
earlgrey/drivers: Port SPI Host driver and add smoke test

lchiawei:
Key modifications and fixes made during the port:
- Implements the `embedded-hal` `SpiDevice` trait, which includes:
  * Defines a related `SpiError` type.
  * Defines the `transaction` method in the struct, which accepts an array of operations, and `csaat` will be asserted during the entire operation series.
  * Marks the operations like `Transfer`, `TransferInPlace`, and `DelayNs` as unsupported explicitly, which is also originally unsupported.
  * Removes unsupported hardware optimization methods like `prefetch`, `drain_fifo`, and `aligned_drain_fifo` since they are not defined in `embedded_hal`. If we need the original optimization that only works on aligned words, we can branch on aligned/unaligned input and apply similar optimization when the input is aligned. However, we should stick with the input type defined in `embedded_hal`.
- Defines a dedicated `SpiConfig` struct and a dedicated `configure` method to make SPI Host configurable.

Also includes a smoke test validating SPI read transactions by checking the SFDP signature.

### DIFF
--- a/target/earlgrey/drivers/spi_host.rs
+++ b/target/earlgrey/drivers/spi_host.rs
@@ -1,0 +1,272 @@
+// Licensed under the Apache-2.0 license
+// SPDX-License-Identifier: Apache-2.0
+
+//! SPI Host driver for Earlgrey.
+//!
+//! This driver implements the `embedded-hal` 1.0 SPI traits for the Earlgrey SPI Host controller.
+
+#![no_std]
+
+use util_error::{
+    ErrorCode, SPI_GENERIC_FIFO_OVERFLOW, SPI_GENERIC_FIFO_UNDERFLOW, SPI_GENERIC_HARDWARE_ERROR,
+    SPI_GENERIC_INVALID_TRANSACTION, SPI_GENERIC_TIMEOUT,
+};
+
+/// SPI Host driver error wrapping an [`ErrorCode`].
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub struct SpiError(pub ErrorCode);
+
+impl SpiError {
+    /// Invalid transaction parameters or state.
+    pub const INVALID_TRANSACTION: Self = Self(SPI_GENERIC_INVALID_TRANSACTION);
+    /// TX FIFO overflow.
+    pub const FIFO_OVERFLOW: Self = Self(SPI_GENERIC_FIFO_OVERFLOW);
+    /// RX FIFO underflow.
+    pub const FIFO_UNDERFLOW: Self = Self(SPI_GENERIC_FIFO_UNDERFLOW);
+    /// Operation timed out.
+    pub const TIMEOUT: Self = Self(SPI_GENERIC_TIMEOUT);
+    /// Hardware error reported by the controller.
+    pub const HARDWARE_ERROR: Self = Self(SPI_GENERIC_HARDWARE_ERROR);
+}
+
+impl embedded_hal::spi::Error for SpiError {
+    fn kind(&self) -> embedded_hal::spi::ErrorKind {
+        // Map all errors to Other for now.
+        embedded_hal::spi::ErrorKind::Other
+    }
+}
+
+impl From<ErrorCode> for SpiError {
+    fn from(err: ErrorCode) -> Self {
+        SpiError(err)
+    }
+}
+
+impl From<SpiError> for ErrorCode {
+    fn from(err: SpiError) -> Self {
+        err.0
+    }
+}
+
+/// Configuration for the SPI Host driver.
+#[derive(Debug, Clone, Copy, Eq, PartialEq)]
+pub struct SpiConfig {
+    /// Clock divider to adjust transfer speed.
+    /// Formula: f_sck = f_core / (2 * (clkdiv + 1))
+    pub clkdiv: u32,
+    /// SPI clock polarity (false for low when idle, true for high when idle).
+    pub cpol: bool,
+    /// SPI clock phase (false to sample on first edge, true on second edge).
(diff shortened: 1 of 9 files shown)


lchiawei:
One pending question is that should we accept a token (current implementation) or a `spi_host::RegisterBlock<ureg::RealMmioMut<'_>>` instead for SPI Host constructor. It seems to me that the usage of token can keep the exclusive ownership more trackable?

target/earlgrey/drivers/spi_host.rs
```diff
@@ -0,0 +1,259 @@
...
+    InvalidTransaction,
+    /// TX FIFO overflow.
+    FifoOverflow,
+    /// RX FIFO underflow.
+    FifoUnderflow,
+    /// Operation timed out.
+    Timeout,
+    /// Hardware error reported by the controller.
+    HardwareError,
+}
```
cfrantz: We should define these in `//util/error` as instances of `ErrorCode`.

cfrantz:
> One pending question is that should we accept a token (current implementation) or a `spi_host::RegisterBlock<ureg::RealMmioMut<'_>>` instead for SPI Host constructor. It seems to me that the usage of token can keep the exclusive ownership more trackable?

In in favor of a single `new` function:
```
impl SpiHost {
    /// Create a new SpiHost instance using the peripheral ownership token for SpiHost0.
    pub fn new(mmio: spi_host::RegisterBlock<ureg::RealMmioMut<'static>>) -> Self {
        Self { mmio }
    }
}
```
In order to have an instance of `SpiHost0` (or 1), you need to invoke `unsafe { SpiHost0::new() }`.  There is nothing that really grants you exclusive access to the the SpiHost registers (you could invoke `new` again), so the added complexity of dealing with the type as a token doesn't really get you anything.  The SpiHost driver doesn't borrow the SpiHost0 or SpiHost1 register block - it consumes it, so I don't really think there is a tracking aspect here (it's `'static`, the lifetime is the lifetime of the program).

I think its more straightforward to create the SpiHost driver using the `PTR` field.  You could make the `SpiHost::new` method `unsafe` and comment that the condition for safe use of the driver is to ensure that you have exclusive access to the peripheral.

lchiawei:
> > One pending question is that should we accept a token (current implementation) or a `spi_host::RegisterBlock<ureg::RealMmioMut<'_>>` instead for SPI Host constructor. It seems to me that the usage of token can keep the exclusive ownership more trackable?
> 
> In in favor of a single `new` function:
> 
> ```
> impl SpiHost {
>     /// Create a new SpiHost instance using the peripheral ownership token for SpiHost0.
>     pub fn new(mmio: spi_host::RegisterBlock<ureg::RealMmioMut<'static>>) -> Self {
>         Self { mmio }
>     }
> }
> ```
> 
> In order to have an instance of `SpiHost0` (or 1), you need to invoke `unsafe { SpiHost0::new() }`. There is nothing that really grants you exclusive access to the the SpiHost registers (you could invoke `new` again), so the added complexity of dealing with the type as a token doesn't really get you anything. The SpiHost driver doesn't borrow the SpiHost0 or SpiHost1 register block - it consumes it, so I don't really think there is a tracking aspect here (it's `'static`, the lifetime is the lifetime of the program).
> 
> I think its more straightforward to create the SpiHost driver using the `PTR` field. You could make the `SpiHost::new` method `unsafe` and comment that the condition for safe use of the driver is to ensure that you have exclusive access to the peripheral.

Done. I applied your suggestions and simplified the constructor of SPI Host.

target/earlgrey/drivers/spi_host.rs
lchiawei: I agreed that we should define the error codes in `util`, and I am thinking about how to implement this - if we define SPI errors as `ErrorCode` instances in util/error, since the host driver must satisfy the `embedded-hal`'s `SpiDevice` trait, the associated error type must implement `embedded_hal::spi::Error`. Are we happy about adding `embedded-hal` as a dependency to `util/error` and implement the trait for `ErrorCode` directly? I am afraid that we are going to add lots of dependencies and `impl`s if we are going to depend on other driver HALs from `embedded_hal` in the future.

cfrantz: We can create a wrapper type to get around the orphan rule and avoid adding a dependency to our error crates.

1. Define our errors in `//util/error`
2. Create the wrapper type and conversion traits:
```
pub struct SpiError(ErrorCode);
impl From<ErrorCode> for SpiError { ... }
impl From<SpiError> for ErrorCode { ... }
```
3. Implement `spi::Error` on the wrapper type.

The conversion traits will allow the `?` operator to automatically convert types in error-exit paths.

lchiawei: Done.

util/error/spi.rs
```diff
@@ -0,0 +1,24 @@
...
+
+use crate::{ErrorCode, ErrorModule};
+use pw_status::Error;
+
+/// The generic SPI error module.
+pub const SPI_GENERIC: ErrorModule = ErrorModule::new(0x5350); // ascii `SP`.
+
+/// Invalid transaction parameters or state.
+pub const SPI_GENERIC_INVALID_TRANSACTION: ErrorCode =
+    SPI_GENERIC.from_pw(0, Error::InvalidArgument);
```
cfrantz: A change a want to make in the future is to avoid using zero as part of the error code.

cfrantz (approved):

util/error/spi.rs
lchiawei: Done.

PR #348, opened 2026-07-15
earlgrey: DFU updates in transport firmware

cfrantz:
- Move the DFU module into the transport firmware and customize is for the transport firmware use case.
- Detect an ownership transfer blob in the firmware payload.
- Sequence an ownership transfer if the ownership blob is present.
- Add tests.

### DIFF
--- a/target/earlgrey/firmware/transport/dfu.rs
+++ b/target/earlgrey/firmware/transport/dfu.rs
@@ -1,0 +1,523 @@
+// Licensed under the Apache-2.0 license
+// SPDX-License-Identifier: Apache-2.0
+
+//! Transport Firmware Device Firmware Upgrade (DFU) handler for Earlgrey.
+//!
+//! This module implements the USB DFU protocol for the Transport Firmware on Earlgrey, supporting firmware
+//! updates (ROM_EXT and Application) and reading device certificates (UDS, CDI0, CDI1).
+
+use earlgrey_sysmgr_client::{BootInfo, SysmgrClient};
+use earlgrey_util::manifest::{
+    Manifest, MANIFEST_EXT_ID_OWNER_TRANSFER_BLOB, MANIFEST_EXT_NAME_OWNER_TRANSFER_BLOB,
+};
+use earlgrey_util::tags::BootSlot;
+use earlgrey_util::tags::ManifestIdentifier;
+use earlgrey_util::EarlgreyFlashAddress;
+use earlgrey_util::PersoCertificate;
+use hal_flash::{Flash, FlashAddress};
+use services_flash_client::FlashIpcClient;
+use util_error::ErrorCode;
+use util_ipc::IpcChannel;
+use zerocopy::FromBytes;
+
+use protocol_usb_dfu::{DfuHandler, DfuStatus};
+use zfmt::Zfmt;
+
+const FLASH_BLOCK_SIZE: usize = 2048;
+
+#[derive(Zfmt)]
+#[zfmt(format = "Flashing {region} region at {start:x}-{end:x}")]
+struct FlashingRegion {
+    region: &'static str,
+    start: u32,
+    end: u32,
+}
+
+#[derive(Zfmt)]
+#[zfmt(format = "Unknown manifest ID: {id:08x}")]
+struct UnknownManifestId {
+    id: u32,
+}
+
+#[derive(Zfmt)]
+#[zfmt(format = "Invalid Application Manifest! Error code: 0x{code:08x}")]
+struct InvalidAppManifest {
+    code: u32,
+}
+
+#[derive(Zfmt)]
+#[zfmt(format = "Found Owner Transfer Blob at offset 0x{offset:x}")]
+struct FoundOwnerTransferBlob {
+    offset: u32,
+}
+
+#[derive(Zfmt)]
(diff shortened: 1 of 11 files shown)


target/earlgrey/firmware/transport/dfu.rs
```diff
@@ -0,0 +1,520 @@
...
+    util_zfmt::debug!("Reading certificate {n}");
+    let (partition, mut n) = match n {
+        0 => (0, 0), // The UDS (dice) cert is located in bank=0, page=9.
+        1 => (1, 0), // The CDI (dice) certs are located in bank=1, page=9.
+        2 => (1, 1), // CDI1 is the second cert in bank=1, page=9.
+        _ => return Err(DfuStatus::ErrFile),
+    };
+    let mut offset = 0usize;
+    let mut buf = [0u8; 1024];
+    loop {
```
anthonychen1251: Would that be better to add an additional check here for `offset` to avoid potential underflow?
```
        if offset >= 2048 {
+            break;
+        }
```

target/earlgrey/firmware/transport/dfu.rs
```diff
@@ -0,0 +1,520 @@
...
+
+    /// Returns the physical flash address offset for a given boot slot.
+    fn addr(slot: BootSlot) -> u32 {
+        match slot {
+            BootSlot::SlotA => 0,
+            BootSlot::SlotB => 0x80000,
+            _ => unreachable!(),
+        }
+    }
+}
```
anthonychen1251, line 245: FYI, [one](https://github.com/OpenPRoT/openprot/commit/ec33cd83fd381dfc85983e248b45aabe255a50ad) of the commits in my [local branch ](https://github.com/OpenPRoT/openprot/compare/earlgrey-hwe...anthonychen1251:openprot:external-flash-fw-update-transport-2)extracts the `FwUpdate` tracker from `target/earlgrey/util/dfu.rs` to an independent library target to be reused in the `updatemgr` process. I will rebase onto your commits and refactor this part once this PR is merged.

target/earlgrey/firmware/transport/dfu.rs
```diff
@@ -0,0 +1,520 @@
...
+    let mut buf = [0u8; 1024];
+    loop {
+        let sz = core::cmp::min(2048 - offset, buf.len());
+        flash
+            .read(
+                FlashAddress::info(partition, 9, offset as u32),
+                &mut buf[..sz],
+            )
+            .map_err(|_| DfuStatus::ErrUnknown)?;
+        match PersoCertificate::from_bytes(&buf) {
```
anthonychen1251: When `sz` is less than 1024 (which happens when `offset > 1024`), the remaining bytes of `buf` contain stale data from the previous iteration.  If the certificate is truncated or corrupted,  `from_bytes` might still succeed by reading the stale data. Could be better to slice the buffer before parsing
```
match PersoCertificate::from_bytes(&buf[..sz]) {
```

target/earlgrey/firmware/transport/dfu.rs
```diff
@@ -0,0 +1,520 @@
...
+                    for entry in &manifest.extensions.entries {
+                        if entry.identifier == MANIFEST_EXT_ID_OWNER_TRANSFER_BLOB {
+                            let offset = entry.offset;
+                            // Ensure the entire fixed part of the extension (Header + 2048 byte blob)
+                            // fits within the reported signed region and overall length.
+                            const EXT_FIXED_SIZE: usize = 8 + 2048;
+                            if offset >= 1024
+                                && (offset as usize + EXT_FIXED_SIZE) <= (manifest.length as usize)
+                                && (offset as usize + EXT_FIXED_SIZE)
+                                    <= (manifest.signed_region_end as usize)
```
anthonychen1251: Since `OWNER_TRANSFER_BLOB` is an unsigned extension, I think it should be placed after  `signed_region_end`. Something might go wrong if this condition is passed when performing the owner transfer test.

anthonychen1251: Ah, just realize this snippet is removed in the following commit. Would be clearer to remove it from this commit.

target/earlgrey/firmware/transport/dfu.rs
cfrantz: Fixed.

target/earlgrey/firmware/transport/dfu.rs
cfrantz: ack.

target/earlgrey/firmware/transport/dfu.rs
cfrantz: Done.

cfrantz: Done

target/earlgrey/firmware/transport/dfu.rs
cfrantz: I'd meant to squash commits together and didn't get to it.

target/earlgrey/firmware/transport/tests/dfu/spec.md
anthonychen1251, line 1: This appears to be a plan note; shall we remove it?

anthonychen1251 (approved):

PR #351, opened 2026-07-15
Pldm Service and Host Tests

CourtneyDrant:
This pull requests attempts to add support for Pldm Service.  Pldm Service includes responder, requester and a firmware device device as a terminus for type 0 and type 5 commands.  Basic connections to the pldm-lib are tested in base_host and firmware_update_host tests.


openprot/README.md
JesseMelon, line 1: This readme is very demo specific, and has citations to documents presumably not available to consumers. Is this readme accomplishing internal planning or a summary of the openprot project to a general consumer? Should it be its own PR?

PR #352, opened 2026-07-16
Add boot reset control

chrysh:
This commit adds the fwmanager service and the BootControl trait to the lib.rs

Furthermore, it renames Monitor to SpiMonitor, so that adding the BootMonitor trait later.


services/fwmanager/api/src/lib.rs
leongross, line 1: We should keep the overall structure of this code base in mind. The firmware manager will have APIs for several components, so maybe we should encapsulate this code into a separate module, such as `boot_control.rs` and keep use `lib.rs` as top level importer for the other modules to come? For reference see structure of `mctp` service: `services/mctp/api/src/`

chrysh: yes, I also noticed that, thanks!

chrysh: I moved it in the latest commit.

chrysh:
I split it into two separate PRs, first one being: https://github.com/OpenPRoT/openprot/pull/355

PR #355, opened 2026-07-17
fwmanager: Add BootControl trait and HAL adapter

chrysh:
Introduce the Boot Orchestrator's actuation capability: the BootControl trait (hold_in_reset / release) and HalBootControl, which binds one HAL ResetControl line to a managed device. Includes a host unit test verifying that holding a device in reset asserts exactly its configured line.

Includes tests that release deasserts the device's configured line and that a controller error surfaces through BootControl unchanged. Extend the fake reset controller with opt-in failure injection to drive the error case.

Closes: https://github.com/9elements/openprot/issues/2

Assisted-by: Claude:claude-opus-4-8

### DIFF
--- a/services/fwmanager/api/src/boot_control.rs
+++ b/services/fwmanager/api/src/boot_control.rs
@@ -3,7 +3,7 @@
 
 //! Actuation capability: hold a managed device in reset and release it.
 
-use openprot_hal_blocking::system_control::ResetControl;
+use openprot_hal_blocking::system_control::{Error as HalError, ErrorKind, ResetControl};
 
 /// Actuation capability: hold a managed device in reset and release it.
 ///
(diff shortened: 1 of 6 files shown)


chrysh:
@rusty1968 @leongross Can you have a look?

services/fwmanager/api/src/boot_control.rs
```diff
@@ -0,0 +1,206 @@
...
+/// store.set_trial(new_slot)?;      // tentative boot selection — not yet committed
+/// dev.release()?;                  // boot the trial image
+/// match monitor.await_boot(window)? {
+///     Booted           => store.commit(new_slot)?,   // observed good => make it active
+///     Failed | Timeout => { /* nothing committed; previous slot still active */ }
+/// }
+/// ```
+pub trait BootControl {
+    /// The error type reported by this device's boot control.
+    type Error: core::fmt::Debug;
```
embediver: Only having `Debug` limits error handling downstream (in the orchestrator).
Requiring `core::error::Error` is more modern as it is available in `core` for a while now.
The openprot-hal implements its own `Error` trait, we might just require that or make the hal compatible with `core::error::Error` (so access to the `ErrorKind` could be acquired by downcasting).

services/fwmanager/api/src/boot_control.rs
```diff
@@ -0,0 +1,206 @@
...
+    fn holding_a_device_in_reset_asserts_its_configured_line() {
+        let mut bmc = HalBootControl::new(MockResetController::new(), BMC_LINE);
+
+        bmc.hold_in_reset().expect("hold_in_reset failed");
+
+        assert_eq!(bmc.controller().calls(), &[Call::Assert(BMC_LINE)]);
+    }
+
+    #[test]
+    fn holding_a_device_in_reset_deasserts_its_configured_line() {
```
embediver:
```suggestion
    fn releasing_a_device_from_reset_deasserts_its_configured_line() {
```

services/fwmanager/api/src/boot_control.rs
```diff
@@ -0,0 +1,206 @@
...
+    #[test]
+    fn controller_error_propagates_through_boot_control() {
+        let mut bmc = HalBootControl::new(
+            MockResetController::failing(ErrorKind::InvalidResetId),
+            BMC_LINE,
+        );
+        let err = bmc
+            .hold_in_reset()
+            .expect_err("expected the controller error to propagate");
+        assert_eq!(err.kind(), ErrorKind::InvalidResetId);
```
embediver: If the `bmc` is a generic `BootControl` calling `err.kind()` won't work.
See the other comment about improving the `BootControl::Error`.
If we change `BootControl` to provide better means to access the error details add a test or change this test to validate it.

services/fwmanager/api/src/boot_control.rs
chrysh: Thanks, very good catch!

services/fwmanager/api/src/boot_control.rs
chrysh: Good point!  `BootControl::Error` now requires `core::error::Error` + `system_control::Error`, so a generic consumer can call `err.kind()` for categorization and gets Display/source() from `core::error::Error`. I also added compile and runtime checks. Does the current iteration make sense to you?

embediver: https://github.com/9elements/openprot/pull/20 removes the HAL dependency from this API.
Adding the HAL error as a trait bound is a bit contradictory.
But I have no strong opinion if we should require `HalError` or not.

services/fwmanager/api/src/boot_control.rs
```diff
@@ -0,0 +1,311 @@
...
+}
+
+/// Adapts any HAL system-control error into a [`core::error::Error`].
+///
+/// Reset controllers keep implementing the HAL `Error`/`kind()` pattern
+/// unchanged; this wrapper supplies the `Display` and `core::error::Error`
+/// machinery [`BootControl::Error`] requires, so no per-implementation work is
+/// needed. The underlying category stays reachable via [`BootError::kind`].
+#[derive(Debug)]
+pub struct BootError<E>(pub E);
```
embediver: Add a `impl<T: HalError> From<T> for BootError<E> {...}`.
This enables implementers to just use the `?` instead of having to call `.map_err()`.

services/fwmanager/api/src/boot_control.rs
chrysh: I moved the boot_control HAL part into services/fwmanager/hal-adapters/src/hal_boot_control.rs, to remove the HAL dependency from the API.

embediver (approved):
LGTM

leongross (approved):

services/fwmanager/api/src/boot_control.rs
rusty1968: +1 on make the hal compatible with core::error::Error

PR #357, opened 2026-07-17
Orchestrator spec and reference impl

rusty1968:


docs/src/design/orchestrator/orchestrator-machine.md
```diff
@@ -0,0 +1,240 @@
...
+
+    AwaitingReady --> AwaitingReady : VerificationPassed [more]\n/ ReleaseReset · ReadFirmware · VerifyFirmware
+    AwaitingReady --> Ready         : ComponentReady [chain done or cursor past end]
+    AwaitingReady --> AwaitingReady : ComponentReady [more]
+    AwaitingReady --> AwaitingReady : VerificationFailed [optional, iRoT pending]
+    AwaitingReady --> Ready         : VerificationFailed [optional, no iRoT pending, chain done]
+    AwaitingReady --> Recovering    : VerificationFailed [required]\n/ RestoreGoldenImage
+
+    state Operational {
+        [*]           --> Ready
```
leongross: Can you elaborate on why this is a valid starting point? I see why `PowerOnReset` starts the process, but I do not get how the prior steps can be skipped to takt this, second entry point?

leongross:
Do you plan to realize your state machine implementation based on https://github.com/typeconstructor/rot-reducer?

docs/src/design/orchestrator/orchestrator-machine.md
rusty1968: I've dropped the `[*] --> Ready` line from the diagram — it wasn't adding information the diagram doesn't already show explicitly elsewhere (`PreSupervision --> Ready` and `AwaitingReady --> Ready`), and removing it
avoids the misreading. misleading as drawn. The only real initial pseudostate is at the top of the diagram:
`[*] --> PowerOnReset`. That's the sole entry — every run starts there

chrysh:
@rusty1968 Should we also add multi-level cascades, concurrent faults, and double-fault recovery tests? Or is this not necessary?

## Answer 
All three are already covered: multi-level cascade (cascading_runtime_corruption_cascades_transitively), concurrent fault (corruption_while_recovering_retargets_to_new_component, corruption_during_update_discards_staged), and double fault (retry_cap_self_latches_via_emit). Happy to add a specific scenario these miss.

services/orchestrator/sm/src/lib.rs
chrysh, line 1: All the other crates have way fewer lines in lib.rs. Would it be better to put it into a separate .rs file?

chrysh:
Are we trying to restore the golden image several times or just once?

## Answer

The state machine isn't aware of the source —  note the effect is named RecoverComponent, not "restore golden image." Golden vs A/B vs last-known-good is the platform driver's call. The reducer only decides how many attempts: up to max_retry per component (set 1 for once).

services/orchestrator/sm/src/lib.rs
```diff
@@ -0,0 +1,1335 @@
...
+                    ctx.emit(Effect::ReleaseReset(*id));
+                    let current_kind = rot.chain.get(rot.cursor as usize).map(|(_, a)| a.kind);
+                    let next_idx = (rot.cursor as usize).saturating_add(1);
+                    if rot.advance_to_next_unheld(ctx, next_idx) {
+                        match current_kind {
+                            Some(ComponentKind::Active) => {
+                                rot.awaiting = Some(*id);
+                                Outcome::Transition(State::AwaitingReady)
+                            }
+                            _ => Outcome::Handled,
```
chrysh, line 527: What if the passive device does not manage to boot up? How will we notice that?

services/orchestrator/sm/src/lib.rs
```diff
@@ -0,0 +1,1335 @@
...
+
+impl Sink {
+    fn new() -> Self {
+        Self {
+            effects: heapless::Vec::new(),
+        }
+    }
+
+    /// Append one effect. Overflow is silently dropped rather than panicking
+    /// (`no_std` safety); overflow means a logic bug.
```
chrysh: Why does overflow mean a logic bug? Silently dropped does not sound like something I want in my security relevant device.

services/orchestrator/sm/src/lib.rs
```diff
@@ -0,0 +1,1335 @@
...
+    fn is_held(&self, id: ComponentId) -> bool {
+        self.held.iter().any(|h| *h == id)
+    }
+
+    /// Advance `cursor` from `start_idx` to the first component not in
+    /// `held`, emitting its `ReadFirmware`/`VerifyFirmware`. Returns `true` if
+    /// found. If the rest of the chain is exhausted or entirely held, sets
+    /// `cursor` to a past-the-end sentinel (`chain.len()`) and returns
+    /// `false` — the caller should treat that as "chain done".
+    fn advance_to_next_unheld(&mut self, ctx: &mut Sink, start_idx: usize) -> bool {
```
chrysh: Device should be in reset when rechecking. A live device is already executing code it loaded before the check, so the recheck says nothing about what's running. It can rewrite its own flash right after the check passes. Either the re-walk re-asserts reset on every non-held component first (recovery = platform re-boot), or re-checking live components becomes an explicitly weaker, differently-named operation instead of the same VerifyFirmware.

services/orchestrator/sm/src/lib.rs
```diff
@@ -0,0 +1,1335 @@
...
+                rot.awaiting = None;
+                let _ = rot.advance_to_next_unheld(ctx, 0);
+            }
+            State::Updating => {
+                ctx.emit(Effect::AuthenticateUpdate);
+                ctx.emit(Effect::StageUpdate);
+            }
+            State::Recovering => {
+                if let Some(failed) = rot.failed {
+                    ctx.emit(Effect::RestoreGoldenImage(failed));
```
chrysh: Why do we always go back to the golden image? Should we try the last known good  before update instead?

docs/src/design/orchestrator/orchestrator-machine.md
```diff
@@ -0,0 +1,411 @@
...
+## Shared storage — `Rot<N>`
+
+Every handler receives a `&mut Rot<N>` alongside the event and the `Sink`. This
+struct is `statig`'s *shared storage*: a single allocation that persists across
+events and is visible to every state and superstate. States carry no data; all
+mutable state lives here.
+
+| Field | Type | Purpose |
+|---|---|---|
+| `chain` | `Vec<(ComponentId, ComponentAttrs), N>` | Ordered trust chain, supplied by the shell at construction time. Never mutated after build. |
```
chrysh: What is a shell in this context? It's easy to confuse with bash or debug shell in the context of firmware. Is there a better word? platform layer vs board layer or so?

services/orchestrator/sm/src/lib.rs
```diff
@@ -0,0 +1,1335 @@
...
+                }
+                Event::UpdateRejected => {
+                    ctx.emit(Effect::DiscardStaged);
+                    Outcome::Transition(State::Ready)
+                }
+                _ => Outcome::Super,
+            },
+
+            State::Recovering => match event {
+                Event::Restored(_) => {
```
chrysh: Is one failed and one retry_count enough for the whole platform? Should it be per device?

chrysh:
Even if we only ever update one device at a time, §5.3.3 background polling can report corruption of device B while device A is mid-update or mid-recovery. What does the machine do then?

chrysh:
Maybe each device should have their own state machine?

services/orchestrator/sm/src/lib.rs
rusty1968: The split is model.rs, tests.rs and lib.rs

services/orchestrator/sm/src/lib.rs
rusty1968: Yes, the retry budget is per component.

rusty1968:
> Even if we only ever update one device at a time, §5.3.3 background polling can report corruption of device B while device A is mid-update or mid-recovery. What does the machine do then?

Excellent question - The scenario is covered by tests: 

- corruption_in_updating_triggers_recovery / corruption_during_update_discards_staged — if a device is found corrupt while another device is being updated, the update is dropped and the machine switches to fixing the corrupt device, throwing away the half-finished update image.

- corruption_while_recovering_retargets_to_new_component / restored_for_wrong_component_does_not_advance_recovery — if a second device fails while we're still fixing the first, we switch to fixing the second; and if the first one later says "I'm fixed," we ignore it, since we'd already moved on.

- cascading_runtime_corruption_cascades_transitively — when the failed device is one we're allowed to simply switch off (rather than one we must stop everything to fix), we isolate it and everything that depends on it, and keep going instead of interrupting the current work.

rusty1968:
> Maybe each device should have their own state machine?

Your intuition is directionally correct — but what genuinely matters here is **per-device state** , and the design already carries it (ComponentStatus, retry_budget, isolation markers,etc).  What stays singular is the ordering: we only fix one device at a time, and we only update one device at a time. The machine keeps a single "who am I fixing right now" slot (State::Recovering(id)), so it can't be pointed at two devices at once. That one-at-a-time behavior is on purpose, not a limitation we ran into.

services/orchestrator/sm/src/lib.rs
rusty1968: Your intuition is sharp here:  the reason we trust golden image is because it is the one trustworthy - it comes from an immutable partition.  Previous could perhaps buy us an availability optimization. When/if we have a stringent availability requirement, we can revisit this, but now perhaps is premature optimization.

rusty1968:
> Do you plan to realize your state machine implementation based on https://github.com/typeconstructor/rot-reducer?

Correct. typeconstructor is another of my github accounts

services/orchestrator/sm/src/lib.rs
chrysh: Advancing the SVN floor before we have a successful boot feels a bit early. If we only advance the SVN when we have a device that booted up, we could go back to the last version of the firmware if the boot fails. 5 years in, the golden image will be very old, so it could be a good idea to jump back to a more recent firmware version first?

chrysh:
> > Maybe each device should have their own state machine?
> 
> Your intuition is directionally correct — but what genuinely matters here is **per-device state** , and the design already carries it (ComponentStatus, retry_budget, isolation markers,etc). What stays singular is the ordering: we only fix one device at a time, and we only update one device at a time. The machine keeps a single "who am I fixing right now" slot (State::Recovering(id)), so it can't be pointed at two devices at once. That one-at-a-time behavior is on purpose, not a limitation we ran into.

What happens if we get another update firmware request for device B while we are fixing the firmware if device A?

services/orchestrator/sm/src/lib.rs
rusty1968: Advancing the SVN floor before we have a successful boot is a real CSA compliance issue. Addressed in c353501023f17e2c66554a269d66d8a0a7785aaa

services/orchestrator/sm/src/lib.rs
rusty1968: CSA boot-progress checkpointing  in 6d5f9564db00e6138181b32c746a6739a9db6742

rusty1968:
> > > Maybe each device should have their own state machine?
> > 
> > 
> > Your intuition is directionally correct — but what genuinely matters here is **per-device state** , and the design already carries it (ComponentStatus, retry_budget, isolation markers,etc). What stays singular is the ordering: we only fix one device at a time, and we only update one device at a time. The machine keeps a single "who am I fixing right now" slot (State::Recovering(id)), so it can't be pointed at two devices at once. That one-at-a-time behavior is on purpose, not a limitation we ran into.
> 
> What happens if we get another update firmware request for device B while we are fixing the firmware if device A?

## TL;DR
The `UpdateRequest` is **silently dropped**. Recovery is not interrupted, no effect is emitted, and nothing is queued. The requester gets no response at all.

That drop is **deliberate as a policy** (recovery outranks update — "fix first, then update"), but **the *silence* is the part worth a second look: the requester cannot tell "refused, busy recovering" apart from "lost."**

## CSA Alignment 

The CSA mandates that failures be reported and models updates as answered requests, which is exactly the gap I am fixing with 57fedb06191810ec20f58a5d4a4280afdbfe2c39

This is a sketch of how we use the report effect.
```rust
impl Platform for PlatformAst  {
    fn execute(&mut self, effect: Effect) -> Result<(), EffectError> {
        match effect {
            // ... device I/O effects (ReadFirmware, ReleaseReset, ...) ...
            Effect::ReportIsolated(id)       => self.mgmt.report_isolated(id),
            Effect::ReportRecoveryFailed(id) => self.mgmt.report_recovery_failed(id),
            // Pure protocol reply — no device I/O. Answer the pending PLDM
            // FW-update request so the requester gets a definite "busy, retry".
            Effect::ReportUpdateDeferred =>
                self.pldm.respond_update(PldmCompletion::RetryLater),
            // ...
        }
    }
}
```

**The CSA does not mandate update-vs-recovery arbitration.** Recovery is described as autonomous and separate from update; the interaction of an in-flight recovery with an incoming update is left to the implementation. Our  _single-flight_, recovery-priority choice is therefore a legitimate    implementation-defined policy, neither required nor forbidden by the CSA.

chrysh:
@rusty1968 Should we add those test?:


```rust
/// INV9: the iRoT gate must hold even when the eRoT walk finishes first.
/// C0's `ComponentReady` is still outstanding when the last verdict
/// arrives; completing the walk must not bypass the gate.
#[test]
fn chain_exhaustion_does_not_bypass_irot_gate() {
    let (_effects, state) = drive(
        chain(&[
            (C0, ComponentAttrs::active_required()),
            (C1, ComponentAttrs::passive_required()),
        ]),
        &[
            BOOT,
            Event::VerificationPassed(C0), // C0 released; its iRoT still owes ComponentReady
            Event::VerificationPassed(C1), // walk done — but C0 never reported
        ],
    );
    assert_eq!(state, State::AwaitingReady(Some(C0)));
}
```


```rust
/// Trust invariant: a verdict only means something for the component under
/// verification. A replayed `VerificationPassed` for a component that has
/// since been isolated must not re-release it.
#[test]
fn replayed_verdict_does_not_release_isolated_component() {
    let (effects, state) = drive(
        chain(&[
            (C0, ComponentAttrs::passive_isolable()),
            (C1, ComponentAttrs::passive_required()),
        ]),
        &[
            BOOT,
            Event::VerificationPassed(C0), // released; walk moves to C1
            Event::CorruptionDetected(C0), // Isolable → AssertReset + Isolated
            Event::VerificationPassed(C0), // replayed / spurious verdict
        ],
    );
    let releases_c0 = effects
        .iter()
        .filter(|&&e| e == Effect::ReleaseReset(C0))
        .count();
    assert_eq!(releases_c0, 1, "isolated component was re-released");
    assert_ne!(state, State::Ready, "C1's verdict is still outstanding");
}
```

```rust
/// A healthy Active component must not be dragged into recovery by a stale
/// boot-progress watchdog. After a re-walk, C0's `ComponentReady` arrives
/// while the machine is back in `PreSupervision`; the report must clear its
/// watchdog (as `Booted` does there), making the later `Timeout` stale.
#[test]
fn component_ready_during_rewalk_clears_watchdog() {
    let (effects, state) = drive(
        chain(&[
            (C0, ComponentAttrs::active_required()),
            (C1, ComponentAttrs::passive_required()),
        ]),
        &[
            BOOT,
            Event::VerificationPassed(C0), // C0 released, watchdog armed
            Event::VerificationFailed(C1), // C1 fails → Recovering(C1)
            Event::Restored(C1),           // restore ok → re-walk from top
            Event::ComponentReady(C0),     // C0's iRoT reports in — discarded
            Event::Timeout(C0),            // stale watchdog fires
        ],
    );
    assert!(
        !effects.contains(&Effect::RestoreGoldenImage(C0)),
        "healthy, ready C0 was sent to recovery by a stale watchdog"
    );
    assert_ne!(state, State::Recovering(C0));
}
```
## Answers

### 1. `chain_exhaustion_does_not_bypass_irot_gate`

**Answer:** We don't block in `AwaitingReady`; a finished walk goes straight to `Ready`, and a released active that never reports is caught by the boot watchdog, not by waiting.

**Covered by: ** `single_active_chain_goes_directly_to_ready`, `timeout_awaited_enters_recovering`

### 2. `replayed_verdict_does_not_release_isolated_component`

**Answer:** A gated component is never re-released: only the component at the cursor can be released, and a replayed verdict for an isolated one is dropped, so `ReleaseReset` fires once.

**Already Covered by: ** `isolable_runtime_corruption_holds_across_rewalk`, `property_verify_before_release_holds_under_random_sequences`

### 3. `component_ready_during_rewalk_clears_watchdog`

**Answer:** A re-walk clears each live component's watchdog via `quiesce_all` on entry, so a stale timeout afterward is dropped and a healthy component is never dragged into recovery.

**Already Covered by: ** `passive_booted_clears_watchdog_then_timeout_is_stale`, `recovery_rewalk_reverifies_live_sibling_at_rest`

docs/src/design/orchestrator/orchestrator-machine.md
```diff
@@ -0,0 +1,348 @@
...
+### `Updating`
+
+An update is in progress.
+
+**Entry action**: emit `AuthenticateUpdate` + `StageUpdate`.
+
+| Event | Guard | Effects | Next state |
+|---|---|---|---|
+| `UpdateVerified` | — | `ActivateUpdate` | `Ready` |
+| `UpdateRejected` | — | `DiscardStaged` | `Ready` (rejected update is not corruption — INV4) |
```
leongross: What does INV4 stand for? Where is this defined?
Also, on the first mention of `rejected` we should have defined what that means in our context. From the CSP, we know that rejection occurs when image verification using DSA fails.

services/orchestrator/sm/README.md
```diff
@@ -0,0 +1,55 @@
...
+recovery).
+
+**No I/O, no hardware.** Every action is an [`Effect`] the surrounding shell
+carries out. Every piece of outside information arrives as an [`Event`].
+
+## Key types
+
+| Type | Role |
+|---|---|
+| `ComponentId` | Opaque `u8` — the shell maps it to hardware; the core never inspects it. |
```
leongross, line 18: `shell` is confusing wording; it is not clear what the shell is at this point in time, reading the documentation. Maybe we can rename it to corpus or harness, or something entirely different.

docs/src/design/orchestrator/orchestrator-model.md
```diff
@@ -0,0 +1,304 @@
...
+
+An opaque `u8` that groups components into a *recovery region*. Components
+assigned the same `RegionId` must be restored together; when any region member
+triggers `Recovering`, the shell issues a joint restore operation for the
+entire region. The core treats the id as an equality key only and never
+inspects the membership.
+
+### `ComponentAttrs`
+
+Per-component attributes that combine three orthogonal axes:
```
leongross: Mentioning 3 orthogonal axes of functionality without explaining it or pointing out their relevance is confusing. What's your intention in writing this?

docs/src/design/orchestrator/orchestrator-machine.md
rusty1968: The invariants are documented only in the tests as docstring, but they should be in the design docs. Fixed in commit c47b868abec6273780b15dfffb3d92748208f69b

rusty1968: The CSA describes two conditions DSA fails and anti-rollback /SVN violation.  I added the definition at first use in 91b14e8a1586065ee29b31bce40e44552062f95a

services/orchestrator/sm/README.md
rusty1968: Shell is confusing - it is meant as the platform-specific driver. which executes the effect. I eliminated the use of the word in e9cf424cc20325548f7d1e14f2671a624bdbb55e

docs/src/design/orchestrator/orchestrator-model.md
rusty1968: This was to describe the grouping in ComponentAttrs but it is confusing. Removed in 42eac7e1c7471e997a35413a70a82cbea2952c71.

docs/src/design/orchestrator/orchestrator-machine.md
rusty1968: I changed it to platform driver.

services/orchestrator/sm/src/lib.rs
rusty1968: This never drops a real effect at runtime — the error branch is impossible to hit. The buffer E is sized at compile time to fit the longest chain:

```rust
const EFFECT_CAP_OK: () = assert!(
    E >= 2 * N + 2,
    "effect buffer E must be >= 2 * chain length N + 2",
);
```

and Rot::new forces that assert, so an under-sized E fails to build:

```rust
pub fn new(chain: Chain<N>, max_retry: u8) -> Self {
    let () = Self::EFFECT_CAP_OK; // won't compile unless E >= 2*N + 2
    ...
}
```

services/orchestrator/sm/src/lib.rs
rusty1968: Agreed — bf06603 implemented exactly your first option: the re-walk re-asserts reset on every non-held component first, so recovery is a platform re-boot.

chrysh (approved):

PR #365, opened 2026-07-24
services: flash client/server 

cfrantz:
This PR depends on #364.

### DIFF
--- a/hal/blocking/flash/driver.rs
+++ b/hal/blocking/flash/driver.rs
@@ -1,0 +1,159 @@
+// Licensed under the Apache-2.0 license
+// SPDX-License-Identifier: Apache-2.0
+
+//! Low-level flash driver interface.
+
+#![no_std]
+
+use core::num::NonZero;
+
+use util_types::PowerOf2Usize;
+use zerocopy::{FromBytes, Immutable, IntoBytes, KnownLayout};
+
+/// Low-level flash driver interface.
+///
+/// This trait defines the interface for interacting with flash hardware at a low level.
+/// It supports an asynchronous-style execution model using a start-poll-complete pattern:
+/// 1. Start the operation (`start_erase`, `start_program`).
+/// 2. Poll status (`is_busy`) or wait for interrupt.
+/// 3. Finalize and check errors (`complete_op`).
+///
+/// Read operations (`read`) are assumed to be synchronous and blocking for simplicity.
+pub trait FlashDriver {
+    /// The error type returned by driver operations.
+    type Error;
+
+    /// The default page size in bytes.
+    const PAGE_SIZE: usize;
+
+    /// The maximum size of a single program operation (write window).
+    /// Program operations cannot span across boundaries aligned to this size.
+    const PROGRAM_WINDOW_SIZE: usize;
+
+    /// The maximum size of a single read operation.
+    const MAX_READ_SIZE: usize;
+
+    /// The alignment required for read operations (addresses and lengths).
+    const READ_ALIGNMENT: usize;
+
+    /// The alignment required for program operations (addresses and lengths).
+    const PROGRAM_ALIGNMENT: usize;
+
+    /// Returns the total size of the flash in bytes.
+    fn size(&self) -> NonZero<usize>;
+
+    /// Returns a bitmap of supported erase block sizes.
+    ///
+    /// Each bit `i` represents a supported erase block size of `2^i` bytes.
+    /// For example, if bit 11 is set, then 2048-byte erases are supported.
+    fn erasable_sizes_bitmap(&mut self) -> Result<u32, Self::Error>;
+
+    /// Reads data from flash synchronously.
+    ///
+    /// # Arguments
+    /// * `start_addr`: The address to start reading from. Must be aligned to `READ_ALIGNMENT`.
+    /// * `buf`: The buffer to read data into. Must have size <= `MAX_READ_SIZE` and be aligned to `READ_ALIGNMENT`.
+    fn read(&mut self, start_addr: FlashAddress, buf: &mut [u8]) -> Result<(), Self::Error>;
+
+    /// Starts an erase operation asynchronously.
+    ///
+    /// The operation is not guaranteed to be complete until `complete_op` is called
+    /// and returns `Ok`.
+    ///
+    /// # Arguments
+    /// * `start_addr`: The start address of the block to erase. Must be aligned to the block size.
+    /// * `size`: The size of the block to erase. Must have its corresponding bit set in `ERASABLE_SIZES_BITMAP`.
+    fn start_erase(
+        &mut self,
+        start_addr: FlashAddress,
+        size: PowerOf2Usize,
+    ) -> Result<(), Self::Error>;
+
+    /// Starts a program operation asynchronously.
+    ///
+    /// The operation is not guaranteed to be complete until `complete_op` is called
+    /// and returns `Ok`.
+    ///
+    /// The programmed region must not cross a `PROGRAM_WINDOW_SIZE` boundary.
+    ///
+    /// # Arguments
+    /// * `start_address`: The address to start programming at. Must be aligned to `PROGRAM_ALIGNMENT`.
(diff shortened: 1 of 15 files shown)


rusty1968:
### The client does not compile

bazelisk build --config=virt_ast10x0 //services/flash/...
error[E0599]: no function or associated item named `check_status` found for struct `ErrorCode` in the current scope
  --> services/flash/client.rs:43:20
   |
43 |         ErrorCode::check_status(result)?;
   |                    ^^^^^^^^^^^^ function or associated item not found in `ErrorCode`
   |
note: if you're trying to build a new `ErrorCode` consider using one of the following associated functions:
      ErrorCode::new
      ErrorCode::kernel_error

That function does not exist anywhere. `util/error/lib.rs:64-79` is the complete inherent impl:

```rust
// util/error/lib.rs:62-79
pub struct ErrorCode(pub NonZero<u32>);

impl ErrorCode {
    pub const fn new(val: u32) -> Self { … }
    pub fn kernel_error(e: pw_status::Error) -> Self { … }
}
```

services/flash/client.rs
```diff
@@ -0,0 +1,109 @@
...
+// SPDX-License-Identifier: Apache-2.0
+
+//! Flash IPC client implementation.
+
+#![no_std]
+use core::num::NonZero;
+
+use hal_flash::{Flash, FlashAddress};
+use services_flash_opcode::*;
+use userspace::time::Instant;
```
rusty1968, line 11: The client crate is reaching around the ipc_util shim to grab `Instant` from the backend directly,
re-introducing the exact dependency the shim was built to hide.

```rust
// services/flash/client.rs
- use userspace::time::Instant;
- use util_ipc::{IpcChannel, IpcHandle};
+ use util_ipc::{IpcChannel, IpcHandle, Instant};
```

services/flash/server.rs
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    /// Handles a single IPC request.
+    ///
+    /// This method performs a non-blocking read on the IPC handle. The caller
+    /// must ensure the handle is readable (e.g., by calling `syscall::object_wait`)
+    /// before calling this method.
+    pub fn handle_one(&mut self, ipc: &IpcHandle, data: &mut [u8]) -> Result<(), ErrorCode> {
+        let len = ipc.read(0, data).map_err(ErrorCode::kernel_error)?;
+        let (opcode, reqrsp) = data.split_at_mut(core::mem::size_of::<Opcode>());
```
rusty1968, line 118: The opcode is read without checking the message is long enough to contain one. 

It is a minor breach for API misuse since data buffer is caller controlled. The server does not defend against a sub-4-byte message. If the client sends fewer than 4 bytes, the opcode is assembled partly from the new message and partly from stale bytes left over from the previous client's request.

services/flash/server.rs
```diff
@@ -0,0 +1,134 @@
...
+    ///
+    /// Parses the `EraseOp` from the input data and erases the specified block.
+    fn handle_erase<'a>(
+        &mut self,
+        data: &'a mut [u8],
+        reqsz: usize,
+    ) -> Result<&'a [u8], ErrorCode> {
+        let req_data = data.get(..reqsz).ok_or(error::IPC_ERROR_BAD_REQ_LEN)?;
+        let op = EraseOp::read_from_bytes(req_data).map_err(|_| error::IPC_ERROR_BAD_REQ_LEN)?;
+        let Some(size) = PowerOf2Usize::new(op.size as usize) else {
```
rusty1968, line 59: The `handle_erase` method only checks if size is a power of two.

It does not check:
-  Whether size is a power-of-2 **the chip actually supports** (the erasable_sizes_bitmap is never read here).
- Whether the address is **aligned** to that size.
- Whether the address is even **inside the device**.

rusty1968:
The overall security guideline we should enforce across all services: Validate untrusted input at the IPC trust boundary, not in the hardware driver implementation.

The flash IPC server is the point where client-controlled address/size/length values first enter the service. Every handler must fully establish the driver's documented preconditions — supported size (bitmap membership), address alignment, and in-bounds range — before calling into BlockingFlash/FlashDriver.

hal/blocking/flash/driver.rs
```diff
@@ -0,0 +1,159 @@
...
+    fn add(self, other: usize) -> Self {
+        Self {
+            offset: self.offset + other as u32,
+        }
+    }
+}
+
+impl core::ops::AddAssign<usize> for FlashAddress {
+    fn add_assign(&mut self, other: usize) {
+        self.offset += other as u32;
```
rusty1968, line 142: This is an uncheked add

hal/blocking/flash/driver.rs
```diff
@@ -0,0 +1,159 @@
...
+    pub fn offset(&self) -> u32 {
+        self.offset
+    }
+}
+
+impl core::ops::Add<usize> for FlashAddress {
+    type Output = Self;
+    fn add(self, other: usize) -> Self {
+        Self {
+            offset: self.offset + other as u32,
```
rusty1968, line 135: This is an uncheked add

hal/blocking/flash/driver.rs
```diff
@@ -0,0 +1,159 @@
...
+    }
+}
+
+impl core::fmt::Debug for FlashAddress {
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        core::fmt::Display::fmt(self, f)
+    }
+}
+
+impl FlashAddress {
```
rusty1968, line 119:
```rust
impl FlashAddress {
    /// Advances the address by `n` bytes, returning `None` on overflow.
    pub fn checked_add(self, n: usize) -> Option<Self> {
        u32::try_from(n).ok()
            .and_then(|n| self.offset.checked_add(n))
            .map(Self::new)
    }
}
```

hal/blocking/flash/flash.rs
```diff
@@ -0,0 +1,430 @@
...
+            self.driver.read(addr, buf_chunk)?;
+            addr += buf_chunk.len();
+        }
+        Ok(())
+    }
+    /// Erases a block of flash.
+    ///
+    /// Starts the asynchronous erase operation and blocks the thread using `self.blocking`
+    /// until the operation completes.
+    fn erase(&mut self, start_addr: FlashAddress, size: PowerOf2Usize) -> Result<(), Self::Error> {
```
rusty1968, line 158: `BlockingFlash` calls `complete_op()` after the wait. 

```rust
self.driver.start_erase(start_addr, size)?;
self.blocking.wait_for_notification();   // returns () — success or timeout indistinguishable
self.driver.complete_op()                // assumes the op finished
```
Even if the wait returned early, the very next line assumes completion and finalizes the operation. There is no branch for "timed out, so quiesce/abort instead." So the control flow, not just the signatures, assumes the wait always succeeds.

services/flash/server.rs
```diff
@@ -0,0 +1,134 @@
...
+            IPC_OP_FLASH_ERASE => self.handle_erase(data, reqsz),
+            IPC_OP_FLASH_PROGRAM => self.handle_program(data, reqsz),
+            IPC_OP_FLASH_READ => self.handle_read(data, reqsz),
+            _ => Err(error::IPC_ERROR_UNKNOWN_OP),
+        }
+    }
+
+    /// Handles a single IPC request.
+    ///
+    /// This method performs a non-blocking read on the IPC handle. The caller
```
rusty1968, line 113: The comments about the ipc read being non-blocking are technically true, but the method handle_one: emphatically blocking. After the read returns, it dispatches through handle_op → handle_erase/handle_program → Flash::erase/program. For the real BlockingFlash, those call self.blocking.wait_for_notification()

rusty1968 (changes requested):
Hi, @cfrantz - Thank you for making this available to us.

cfrantz:
> Hi, @cfrantz - Thank you for making this available to us.

Thank you for the thorough review.  I will try to address your comments in the next couple of days.

rusty1968:
The flash service binds one server to one device. But several flash devices can share a single-owner steering resource that serves one device at a time. Independent servers would race on it, so they can't run as peers. A single server must arbitrate all of them — which requires a per-request device selector so one channel can address each device safely. This is the case for the ASPEED ast1060. There are downsides to this, so we don't need to explore this right away.

PR #367, opened 2026-07-24
fwmanager: Add device-table schema; declare a mock board in target/mock

chrysh:
fwmanager-api gains the schema for the per-board device table: DeviceConfig (name, reset line, boot timeout), CommitPolicy, and a const validate() so a bad table fails the build. All values are board config and live in the new target/mock/devices.rs — a mock board with one device per archetype (direct-flash SPI "bmc", PLDM "nic") whose declaration order is the boot order; hardware targets declare their own tables later. Kept out of config.rs, which belongs to the kernel.

Part of 9elements/openprot#1 (the downstream device definitions; partitioning schemes and verification keys remain open there).

### DIFF
--- a/services/fwmanager/api/src/config.rs
+++ b/services/fwmanager/api/src/config.rs
@@ -1,0 +1,41 @@
+// Licensed under the Apache-2.0 license
+// SPDX-License-Identifier: Apache-2.0
+
+//! Schema for the per-board device table. Board device tables
+//! (`target/<board>/devices.rs`) declare the values; no concrete line or
+//! device is named here.
+
+/// What the orchestrator requires before it commits a staged image.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum CommitPolicy {
+    /// The device reports it came up.
+    Liveness,
+    /// Liveness plus SPDM re-attestation of the running image.
+    LivenessAndAttestation,
+}
+
+/// One managed downstream device, as declared by the board config.
+#[derive(Debug, Clone, Copy)]
+pub struct DeviceConfig {
+    pub name: &'static str,
+    /// Reset line id, passed to HalBootControl::new.
+    pub reset_line: u8,
+    /// How long the orchestrator waits for this device to report Booted
+    /// before it declares a timeout.
+    pub boot_timeout: core::time::Duration,
+    pub commit_policy: CommitPolicy,
+}
+
+/// Checks a device table. Board configs call this in a const context so a
+/// bad table fails the build.
+pub const fn validate(devices: &[DeviceConfig]) {
+    let mut i = 0;
+    while i < devices.len() {
+        assert!(!devices[i].name.is_empty(), "device name must not be empty");
+        assert!(
+            !devices[i].boot_timeout.is_zero(),
+            "boot timeout must not be zero"
+        );
+        i += 1;
+    }
+}
--- a/services/fwmanager/api/src/config.rs
+++ b/services/fwmanager/api/src/config.rs
@@ -6,6 +6,11 @@
 //! device is named here.
 
 /// What the orchestrator requires before it commits a staged image.
+///
+/// Intentionally exhaustive (not `#[non_exhaustive]`): adding a variant is
+/// a breaking change, so the compiler forces every match on the policy —
+/// in particular the orchestrator's commit decision — to handle the new
+/// variant explicitly instead of falling into a wildcard arm.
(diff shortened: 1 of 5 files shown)


chrysh:
@FerralCoder @rusty1968 @embediver @leongross Can you have a look at a possible target config for the attached downstream and upstream devices?

services/fwmanager/api/src/config.rs
```diff
@@ -0,0 +1,41 @@
...
+    /// Liveness plus SPDM re-attestation of the running image.
+    LivenessAndAttestation,
+}
+
+/// One managed downstream device, as declared by the board config.
+#[derive(Debug, Clone, Copy)]
+pub struct DeviceConfig {
+    pub name: &'static str,
+    /// Reset line id, passed to HalBootControl::new.
+    pub reset_line: u8,
```
rusty1968: Can we make the schema opaque/generic over the id, or at least document that it targets only ResetId = u8 controllers? DeviceConfig::reset_line is a u8, but HalBootControl is generic over C::ResetId — so "passed to HalBootControl::new" only type-checks when ResetId == u8, which the adapter's own test comment says won't hold for a concrete driver ("Real driver should use Enum"). Nothing enforces that the u8 matches the controller's actual id type; it's only stated in a comment, so a mismatch fails at the call site (or silently misbehaves) instead of being caught by the type system.

services/fwmanager/api/src/config.rs
```diff
@@ -0,0 +1,41 @@
+// Licensed under the Apache-2.0 license
+// SPDX-License-Identifier: Apache-2.0
+
+//! Schema for the per-board device table. Board device tables
+//! (`target/<board>/devices.rs`) declare the values; no concrete line or
+//! device is named here.
+
+/// What the orchestrator requires before it commits a staged image.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
```
rusty1968, line 14: Consider marking public enums/structs #[non_exhaustive] so they can be extended without breaking downstream callers - unless callers must construct or match them by hand. In that case, document on the type that it's intentionally exhaustive and that adding fields/variants is a breaking change

rusty1968 (changes requested):

services/fwmanager/api/src/config.rs
FerralCoder: i'd recommend using a more flexible type for the reset identifier. It may not always be as simple as assuming that an integer is enough to identify the reset control. Maybe there is a platform with multiple GPIO controllers and you have to specify the controller number and pin number. Maybe there is a system where a different mechanism is used for reset control (say passing a message to another device). Same for other mechanisms, keep it simple while allowing for a way to handle unknowns that may exist in some (future) platform type

services/fwmanager/api/src/config.rs
rusty1968: There are two pub enums in this crate that should be#[non_exhaustive]

services/fwmanager/api/src/config.rs
```diff
@@ -0,0 +1,41 @@
...
+
+/// One managed downstream device, as declared by the board config.
+#[derive(Debug, Clone, Copy)]
+pub struct DeviceConfig {
+    pub name: &'static str,
+    /// Reset line id, passed to HalBootControl::new.
+    pub reset_line: u8,
+    /// How long the orchestrator waits for this device to report Booted
+    /// before it declares a timeout.
+    pub boot_timeout: core::time::Duration,
```
FerralCoder: liveness may be more complex than this. consider that there may be multiple boot progress checkpoints, each with its own timeout/watchdog value. Consider that it may not be a positive signal sent by the device but may require OpenPRoT to poll on an I2C address, poll on a register, send a "ping" type message and look for response, etc....  flexibility/extensibility is key. We will only build the simple solution we need today, but plan for the flexibility to handle something more complex.

FerralCoder (commented):
am I correct to assume this is just a starting point showing the concept, and that more config parameters are intended to be added in future commits?

services/fwmanager/api/src/config.rs
embediver: +1 for making the `reset_line` type generic.

chrysh: Good catch!

rusty1968 (approved):

rusty1968:
Possible follow ups: https://github.com/OpenPRoT/openprot/issues/375, #377

PR #368, opened 2026-07-24
fwmanager: Add BootMonitor capability and GpioBootMonitor adapter

chrysh:
BootMonitor reports a device's boot liveness as BootStatus in the api leaf crate; GpioBootMonitor implements it over one HAL GpioPort input line. MonitorError meets the core::error::Error bound and keeps the concrete HAL error reachable through source(). new() rejects empty pin masks; boot latches are cleared by the reset path, never the observer.

Closes 9elements/openprot#6.

(Replaces 9elements/openprot#20, which was opened against the fork's main by mistake.)

### DIFF
--- a/services/fwmanager/api/src/boot_monitor.rs
+++ b/services/fwmanager/api/src/boot_monitor.rs
@@ -1,0 +1,175 @@
+// Licensed under the Apache-2.0 license
+// SPDX-License-Identifier: Apache-2.0
+
+//! Observation capability: read a managed device's boot liveness.
+
+/// Liveness of a managed device's boot: Boot Confirmation only.
+///
+/// Reports only that a device came up, never what booted; confirming the
+/// running image is the one the RoT staged is attestation, a separate step.
+/// `Failed` is optional device-reported evidence and never the only failure
+/// path, since a hung device reports nothing — a stuck boot is caught by the
+/// orchestrator's timeout, not by this enum.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum BootStatus {
+    /// Held in reset by the orchestrator; not yet released.
+    InReset,
+    /// Released, but boot completion not yet observed.
+    Booting,
+    /// Boot completion observed.
+    Booted,
+    /// Device reported a boot failure.
+    Failed,
+}
+
+/// Observation capability: read a managed device's boot liveness.
+///
+/// Pull-shaped: where the underlying signal is an edge or pulse, the interrupt
+/// latches a flag beneath this seam and `boot_status` only reads it. No
+/// callback registration, which would require allocation and invert control
+/// into device implementations.
+///
+/// The reported status must describe the **current** boot cycle. An
+/// implementation backed by a latched signal must guarantee the latch is
+/// cleared whenever the device re-enters reset, so evidence left over from a
+/// previous boot never reads as [`BootStatus::Booted`]. This trait
+/// deliberately has no re-arm operation: clearing is the reset path's job
+/// (hardware tying the latch to the device's reset line, or the same platform
+/// code that drives `BootControl`), not the observer's — a monitor that could
+/// clear its own evidence would let a read race a reset.
+pub trait BootMonitor {
+    /// The error type reported by this device's boot monitor.
+    ///
+    /// Requires [`core::error::Error`] (in `core` since Rust 1.81) so the
+    /// orchestrator gets `Display` and a `source()` cause chain, not just a
+    /// `Debug` dump. Error categories stay implementation-defined — this
+    /// crate names no error vocabulary of its own; a consumer that knows the
+    /// concrete adapter can recover its details by downcasting the
+    /// `&dyn core::error::Error`.
+    type Error: core::error::Error;
+
+    /// Returns the current liveness of the device.
+    ///
+    /// # Errors
+    ///
+    /// Returns an error if the underlying liveness signal cannot be read.
+    fn boot_status(&self) -> Result<BootStatus, Self::Error>;
+}
+
+#[cfg(test)]
+#[allow(clippy::bool_assert_comparison)]
+mod tests {
+    use super::*;
+    use core::cell::Cell;
+
+    // ── Trait contract ──────────────────────────────────────────────────
+    // MockMonitor implements the trait without any HAL dependency. If a
+    // HAL-specific bound sneaks back onto `Error`, this module stops
+    // compiling.
+
+    struct MockMonitor {
+        ready_after: usize,
+        polls: Cell<usize>,
+        fail: bool,
+    }
+
+    #[derive(Debug, PartialEq, Eq)]
+    struct MockFault;
+
+    impl core::fmt::Display for MockFault {
+        fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+            f.write_str("mock monitor fault")
+        }
+    }
+
+    impl core::error::Error for MockFault {}
+
+    impl BootMonitor for MockMonitor {
+        type Error = MockFault;
+
+        fn boot_status(&self) -> Result<BootStatus, MockFault> {
+            if self.fail {
+                return Err(MockFault);
+            }
+            let polls = self.polls.get();
+            self.polls.set(polls + 1);
+            Ok(if polls >= self.ready_after {
+                BootStatus::Booted
+            } else {
+                BootStatus::Booting
+            })
+        }
+    }
+
+    // A device that is still coming up reads Booting, then Booted once it
+    // is up.
+    #[test]
+    fn status_progresses_from_booting_to_booted() {
+        let mon = MockMonitor {
+            ready_after: 1,
+            polls: Cell::new(0),
+            fail: false,
+        };
+
+        assert_eq!(
+            mon.boot_status().expect("boot_status failed"),
+            BootStatus::Booting
+        );
+        assert_eq!(
+            mon.boot_status().expect("boot_status failed"),
+            BootStatus::Booted
+        );
+    }
+
+    #[test]
+    fn errors_surface_through_the_generic_seam() {
+        let mon = MockMonitor {
+            ready_after: 0,
+            polls: Cell::new(0),
+            fail: true,
+        };
+
+        let err = comes_up_within(&mon, 1).expect_err("expected the monitor fault");
+
+        // Display comes from the core::error::Error bound, not a Debug dump.
+        assert_eq!(err.to_string(), "mock monitor fault");
+    }
+
+    // ── The orchestrator's future shape ─────────────────────────────────
+    // Usage examples for the future orchestrator, not API guarantees; move
+    // these to the orchestrator crate once it exists.
+
+    /// Poll a monitor up to `poll_budget` times. `Booting` is not a failure;
+    /// `Ok(false)` means the budget ran out before the device came up.
+    fn comes_up_within<M: BootMonitor>(mon: &M, poll_budget: usize) -> Result<bool, M::Error> {
+        for _ in 0..poll_budget {
+            if mon.boot_status()? == BootStatus::Booted {
+                return Ok(true);
+            }
+        }
+        Ok(false)
+    }
+
+    // A device that comes up within the poll budget reads Booted.
+    #[test]
+    fn a_device_that_comes_up_within_budget_is_booted() {
+        let mon = MockMonitor {
+            ready_after: 2,
+            polls: Cell::new(0),
+            fail: false,
+        };
+
+        assert_eq!(comes_up_within(&mon, 5).expect("boot_status failed"), true);
+    }
+
+    #[test]
+    fn a_device_that_never_comes_up_is_a_timeout_not_an_error() {
+        let mon = MockMonitor {
+            ready_after: usize::MAX,
+            polls: Cell::new(0),
+            fail: false,
+        };
+
+        assert_eq!(comes_up_within(&mon, 3).expect("boot_status failed"), false);
+    }
+}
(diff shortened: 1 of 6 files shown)


services/fwmanager/api/src/boot_monitor.rs
```diff
@@ -0,0 +1,175 @@
...
+
+/// Liveness of a managed device's boot: Boot Confirmation only.
+///
+/// Reports only that a device came up, never what booted; confirming the
+/// running image is the one the RoT staged is attestation, a separate step.
+/// `Failed` is optional device-reported evidence and never the only failure
+/// path, since a hung device reports nothing — a stuck boot is caught by the
+/// orchestrator's timeout, not by this enum.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum BootStatus {
```
embediver, line 14: Lets make this `#[non_exhaustive]`.

Also, do we need an `Unknown` variant?

chrysh: non_exhaustive: Don't we want every consumer to get a compile error when a value is added, so that they must consciously decide how to handle the new state. non_exhaustive forces consumers to write a _ arm, and since every BootStatus variant is decision-relevant for the orchestrator, a future variant silently falling into that arm seems worse than a breaking change.

chrysh: Same reasoning for Unknown: a hung device reports nothing, so "unknown" is already covered by the timeout path (Booting until the orchestrator gives up). It would hide a bad state.

Do you have strong opinions for either?

embediver: No strong opinions. 
I'm fine with unknown always represented by the error path.
Maybe AMD folks have an opinion whether it should be non_exhaustive or not.

services/fwmanager/hal-adapters/src/gpio_boot_monitor.rs
```diff
@@ -0,0 +1,350 @@
...
+    /// # Errors
+    ///
+    /// Propagates any error returned by the port's `read_input`.
+    fn boot_status(&self) -> Result<BootStatus, Self::Error> {
+        let high = self.port.read_input()?.contains(self.ready_pin);
+        let booted = match self.active {
+            ActivePolarity::ActiveHigh => high,
+            ActivePolarity::ActiveLow => !high,
+        };
+        Ok(if booted {
```
rusty1968, line 146: `BootStatus` has four values: `InReset`, `Booting`,
`Booted`, and `Failed`. But this monitor only looks at a single wire and asks
one question — "is it on or off?" — so it can only ever answer `Booting` or
`Booted`:

```rust
Ok(if booted { BootStatus::Booted } else { BootStatus::Booting })
```

It never returns `InReset` or `Failed`. A device that's still held in reset looks exactly the same on that one wire as a device that's been let go but hasn't finished starting up — both read as `Booting`. And there's no way for a device to say "I tried to boot and failed" over a single ready wire, so `Failed` never comes up either.  A consumer would assume from the return type that this API returns the wide enum.

services/fwmanager/api/src/boot_monitor.rs
```diff
@@ -0,0 +1,175 @@
...
+    /// Held in reset by the orchestrator; not yet released.
+    InReset,
```
rusty1968: `BootStatus` is the little list of values a boot monitor can report about a device: `InReset`, `Booting`, `Booted`, `Failed`. That said `InReset` doesn't belong on that list. It's not something a monitor can see — it's something the orchestrator already did. The list should just be `Booting`, `Booted`, `Failed`.

services/fwmanager/api/src/boot_monitor.rs
rusty1968: > non_exhaustive: Don't we want every consumer to get a compile error when a value is added, so that they must consciously decide how to handle the new state. non_exhaustive forces consumers to write a _ arm, and since every BootStatus variant is decision-relevant for the orchestrator, a future variant silently falling into that arm seems worse than a breaking change.

I thought for a bit and exhaustive is right call here. Adding another variant should be a build error. The change should be one atomic commit that touches not only the enum but every consumer the compiler flags, reviewed together in a single PR.

services/fwmanager/hal-adapters/src/gpio_boot_monitor.rs
chrysh: That's not wrong. How would you solve this issue best? Make it return a bool?

services/fwmanager/api/src/boot_monitor.rs
```diff
@@ -0,0 +1,175 @@
...
+    /// Requires [`core::error::Error`] (in `core` since Rust 1.81) so the
+    /// orchestrator gets `Display` and a `source()` cause chain, not just a
+    /// `Debug` dump. Error categories stay implementation-defined — this
+    /// crate names no error vocabulary of its own; a consumer that knows the
+    /// concrete adapter can recover its details by downcasting the
+    /// `&dyn core::error::Error`.
+    type Error: core::error::Error;
+
+    /// Returns the current liveness of the device.
+    ///
```
rusty1968, line 50: Document this only returns a subset.

    /// Any given monitor may only ever produce a *subset* of [`BootStatus`],
    /// depending on the signals it can access: a single ready pin yields only
    /// `Booting`/`Booted`, while a fault-channel backend can also report
    /// `Failed`. This is a capability difference between backends, not an
    /// incomplete implementation. Consumers must still handle the full set —
    /// they cannot know statically which backend they hold.

services/fwmanager/hal-adapters/src/gpio_boot_monitor.rs
rusty1968: This can be solved with the rustdoc so the reader knows the intent. It is not a bug it is just for clarity.

services/fwmanager/hal-adapters/src/gpio_boot_monitor.rs
```diff
@@ -0,0 +1,350 @@
...
+    /// contract offers no way to reject them here, so they are unspecified
+    /// behavior, not a feature).
+    ///
+    /// # Panics
+    ///
+    /// Panics if `ready_pin` is empty. An empty mask is vacuously contained
+    /// in every sample, so it would report every device `Booted` from the
+    /// moment reset is released — a misbinding that must fail at
+    /// construction, in platform bring-up, not stay silent in the field.
+    pub fn new(port: P, ready_pin: P::Mask, active: ActivePolarity) -> Self {
```
rusty1968: When you build a `GpioBootMonitor`, it takes the GPIO port and keeps it for itself — and that "port" is a whole *bank* of pins, not the single pin the
monitor actually watches. So the first monitor you build swallows the entire bank, and there's nothing left for a second one.

That's a problem whenever two devices put their boot-complete signals on different pins of the *same* bank, which is a normal way to wire a board. You'd
naturally want one monitor per device, but you can't: after the first `GpioBootMonitor::new`, the bank is gone.

**It's not hypothetical.** Our immediate target, the AST1060 SGPIOM, is a  serial GPIO expander whose `P` is `SgpiomBankPort` — a whole 32-line register
  bank, not a single pin. Because 32 unrelated lines are packed into each fixed bank, different devices' ready lines routinely land in the *same* bank by
construction.

services/fwmanager/hal-adapters/src/gpio_boot_monitor.rs
```diff
@@ -0,0 +1,350 @@
...
+/// latched in hardware at bring-up so `read_input` still reflects it.
+///
+/// Where a hardware latch is used, the platform must clear it whenever the
+/// device re-enters reset (typically by wiring the latch's clear to the
+/// device's reset line) — [`BootMonitor`] requires that evidence from a
+/// previous boot never reads as [`BootStatus::Booted`], and this adapter only
+/// reads the line, it cannot re-arm it.
+///
+/// [`HalBootControl`]: crate::HalBootControl
+pub struct GpioBootMonitor<P: GpioPort> {
```
rusty1968: Consider borrowing the port instead of owning it.

Make the monitor hold a shared reference:

```rust
pub struct GpioBootMonitor<'a, P: GpioPort> {
    port: &'a P,
    ready_pin: P::Mask,
    active: ActivePolarity,
}
```

Because `read_input(&self)` already needs only `&self`, nothing else in the monitor has to change. Several monitors can now borrow the same bank at once:

```rust
let mon_a = GpioBootMonitor::new(&bank, pin3, ActiveHigh);
let mon_b = GpioBootMonitor::new(&bank, pin5, ActiveHigh);
let mon_c = GpioBootMonitor::new(&bank, pin7, ActiveHigh);
```
This directly matches the read-only nature of the adapter; removes the  single-owner constraint with the smallest possible change.

The issue is it Introduces a lifetime parameter on the type, so the platform must   keep the bank alive somewhere for at least as long as the monitors.

What is your take?

services/fwmanager/hal-adapters/src/gpio_boot_monitor.rs
chrysh: Good point. Fixed.

services/fwmanager/hal-adapters/src/gpio_boot_monitor.rs
chrysh: That's a very good point. Shared &P only works because monitors read. If the same bank also carries output lines it will need &mut P. In this case the &mut of the controller and the & the monitor borrows conflict. In case this happens, we will have to do what SplitPort is doing.

services/fwmanager/api/src/boot_monitor.rs
chrysh: Agreed

rusty1968 (approved):

leongross (approved):
lgtm

services/fwmanager/api/src/boot_monitor.rs
```diff
@@ -0,0 +1,180 @@
...
+/// orchestrator's timeout, not by this enum.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum BootStatus {
+    /// Released, but boot completion not yet observed.
+    Booting,
+    /// Boot completion observed.
+    Booted,
+    /// Device reported a boot failure.
+    Failed,
+}
```
FerralCoder, line 21: additional states I think we need
* "reset" - have not started booting this device, still held in reset

these next two are needed, but may belong elsewhere 
* "authenticated" - device firmware has been retrieved and signature has been verified
* "measured" - device firmware has been measured

these would all come before the "booting" state in this list

rusty1968: > additional states I think we need
> 
> * "reset" - have not started booting this device, still held in reset
> 
> these next two are needed, but may belong elsewhere
> 
> * "authenticated" - device firmware has been retrieved and signature has been verified
> * "measured" - device firmware has been measured
> 
> these would all come before the "booting" state in this list

The criteria to add a variant here is that it should be something that device can report. : "authenticated" and "measured" — RoT-side facts, not device signals. These are produced by the RoT, not reported by the device.

InReset as per the previous discussion is control plane  not observation and it was removed.

rusty1968:
## API usage model
Consider this bare metal toy example just for illustration purposes. 

```rust
use fwmanager_api::{BootMonitor, BootStatus};
use orchestrator_sm::{ComponentId, Event, Orchestrator, Platform};

/// The shell's per-component boot watch. Called after the SM has emitted
/// `Effect::ReleaseReset(id)` for `id` and the platform actually let it go.
///
/// This is the *only* place `BootStatus` crosses into the SM: a `Booted` read
/// becomes `Event::Booted`, a device-reported `Failed` short-circuits into
/// recovery, and "still `Booting` when the budget runs out" is the watchdog
/// firing `Event::Timeout`. Reset / authenticate / measure never appear here —
/// the SM already drove those before it emitted `ReleaseReset`.
fn watch_boot<const N: usize, const E: usize, M, P>(
    sm: &mut Orchestrator<N, E>,
    platform: &mut P,
    monitor: &M,
    id: ComponentId,
    poll_budget: usize,
) where
    M: BootMonitor,
    P: Platform,
{
    for _ in 0..poll_budget {
        match monitor.boot_status() {
            // Came up: hand the SM its post-release liveness event. The SM
            // clears this component's boot-progress watchdog and advances the
            // chain walk (or, for an Active part, this pairs with ComponentReady).
            Ok(BootStatus::Booted) => {
                sm.dispatch(platform, Event::Booted(id));
                return;
            }

            // Released but not up yet — not a failure. Keep waiting inside the
            // budget; the SM is unaware a poll even happened.
            Ok(BootStatus::Booting) => continue,

            // Device reported a fault. The SM has no dedicated "boot failed"
            // event — a device that doesn't come up enters recovery via the
            // boot-progress watchdog — so surface an early, explicit fault as
            // that same Timeout path instead of waiting the window out.
            Ok(BootStatus::Failed) => {
                sm.dispatch(platform, Event::Timeout(id));
                return;
            }

            // Couldn't even read the signal. Treat an unreadable liveness
            // channel as no-progress and let recovery decide.
            Err(_e) => {
                sm.dispatch(platform, Event::Timeout(id));
                return;
            }
        }
    }

    // Budget exhausted, never saw Booted: the boot-progress watchdog fires.
    // The SM routes Timeout(id) for an awaiting component straight into recovery.
    sm.dispatch(platform, Event::Timeout(id));
}
```

services/fwmanager/api/src/boot_monitor.rs
FerralCoder: makes sense, thanks Anthony

FerralCoder (approved):

PR #369, opened 2026-07-25
Error cleanups

cfrantz:
1. Add conversion functions for pigweed errors and other ErrorKinds.
2. Allow inspecting and comparing ErrorModule.
3. Create an `earlgrey_util_error` crate and adjust existing code to use that crate.
4. Rename and move `FLASH_OPENTITAN` to `EG_FLASH` in earlgrey_util_error.


util/error/README.md
```diff
@@ -21,35 +21,44 @@ pub const MY_MODULE: ErrorModule = ErrorModule::new(0x4d59);
 
 An `ErrorCode` is a 32-bit value composed of:
 *   Upper 16 bits: The `ErrorModule` ID.
-*   Lower 16 bits: A module-specific error value.
+*   Lower 16 bits: A module-specific error value (which may be further partitioned).
 
 `ErrorCode` implements `core::error::Error`, `Display`, and `Debug`. It formats as a hex representation of the 32-bit value (e.g., `0x4b450001`).
 
+You can extract the module ID using the [`module()`](lib.rs#L89) method.
```
lchiawei: Nit: I am not sure if we would like to add code link to a specific line in the readme because it will get outdated soon.

anthonychen1251 (approved):

util/error/lib.rs
```diff
@@ -50,9 +55,21 @@ impl ErrorModule {
...
+    /// - Bits 0..7: The Pigweed `err` status (normally fits in 5 bits)
+    ///
+    /// # Panics
+    /// Panics if `code` is zero.
+    pub const fn from_pw(self, code: u8, err: pw_status::Error) -> ErrorCode {
+        // pw_status::Error is 5 bits, but we shift the error code up 8 bits
+        // so the hex representation is easy to read..
+        match NonZero::new(code as u16) {
+            Some(val) => self.error((val.get() << 8) | (err as u16)),
+            None => panic!("Error `code` must be non-zero"),
```
lchiawei, line 71: Just want to understand a little more here (you can directly resolve this comment):

* For error defined as a constant with val = 0, it will definitely causes a panic. However, panic caused during constant evaluation in compilation time will turn into a compilation error.
* For error defined as runtime variable, if the compiler can not prove it will always be 0, then the ELF will contain the symbol of panic and it will fail the `rust_binary_no_panics_test`

Is my understanding correct?

lchiawei (approved):

util/error/README.md
cfrantz: Fixed.

util/error/lib.rs
cfrantz: That's correct.  It is expected that this function will only be used in `const` contexts.

PR #370, opened 2026-07-25
util_panic: clean up panic handlers

cfrantz:


anthonychen1251 (approved):

lchiawei:
Ah I realized that we may need to fix the SPI tests as well: #374

PR #371, opened 2026-07-26
earlgrey: refactor gpio and pinmux drivers 

cfrantz:
Refactor the gpio and pinmux drivers to use pinmux peripheral and pad related constants from `top_earlgrey`.

This PR depends on #369.  You need only review the last commit.

PR #378, opened 2026-07-28
earlgrey: Implement initial platform service 

cfrantz:
Implement the initial version of the platform service.  Tests will be added in follow-up PRs.

This PR depends on #371 and #372.  You only need to review the last 2 commits.

### DIFF
--- a/target/earlgrey/firmware/hwe/platform.rs
+++ b/target/earlgrey/firmware/hwe/platform.rs
@@ -5,19 +5,136 @@
 #![no_main]
 
 use pw_status::Error;
-use userspace::time::{sleep_until, Clock, Duration, SystemClock};
 use userspace::{process_entry, syscall};
 use util_error::{AsStatus, ErrorCode};
 use util_zfmt::messages::{ProcessExit, ProcessStart};
 
-/*
- * TODO: implement platform server.
- */
-
 fn platform_server() -> Result<(), ErrorCode> {
+    use earlgrey_gpio::EarlGreyGpio;
+    use earlgrey_pinout::dualsbs::DualSideBySide;
+    use earlgrey_pinout::swstraps::SwStraps;
+    use earlgrey_pinout::Pinout;
+    use earlgrey_platform::reset::{ResetPolicy, TargetCpuReset};
+    use earlgrey_platform::server::PlatformServer;
+    use earlgrey_platform::spimux::SpiMuxHandler;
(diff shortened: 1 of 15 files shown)


lchiawei (approved):

target/earlgrey/firmware/hwe/platform.rs
```diff
@@ -5,19 +5,135 @@
...
+                    DualSideBySide::RST_CTRL0_N,
+                    DualSideBySide::RST_MON0_N,
+                    DualSideBySide::RST_MON1_N,
+                )),
+                signals::GPIO_16,
+                signals::GPIO_17,
+                signals::GPIO_18,
+            )
+        }
+        _ => {
```
anthonychen1251, line 69: nit: Would that be better to combine them into a single wild card arm e.g. `0..=2 | _ => ...` for now to eliminate code duplication?

target/earlgrey/firmware/hwe/platform.rs
```diff
@@ -5,19 +5,135 @@
...
+    use earlgrey_platform::server::PlatformServer;
+    use earlgrey_platform::spimux::SpiMuxHandler;
+    use earlgrey_platform::usbmux::UsbMuxHandler;
+    use earlgrey_sysmgr_client::{ResetInfo, SysmgrClient};
+    use platform_codegen::{handle, signals};
+    use userspace::syscall::Signals;
+    use userspace::time::{Clock, Duration, SystemClock};
+    use util_ipc::IpcHandle;
+
+    let mut gpio = unsafe { EarlGreyGpio::new() };
```
anthonychen1251, line 27: nit: safety comment

target/earlgrey/firmware/hwe/platform.rs
cfrantz: I changed the values to named constants.  I want to keep the `_` case so that its always there as a catch-all regardless of other configurations.

target/earlgrey/firmware/hwe/platform.rs
cfrantz: Fixed.

target/earlgrey/firmware/hwe/platform.rs
anthonychen1251: Ack.

anthonychen1251 (approved):

PR #379, opened 2026-07-28
Pldm improvement

CourtneyDrant:
This pull request was done to improve on a previous design of the pldm-service.  It has one process which is the Firmware Device.  Firmware Device is the OpenProt Pldm terminus for FW update.  It will loop to receive messages from the UA and when ready, will request data from the UA.  The loop ensures that between each request the terminus will check for UA requests to receive status or cancel an update.


services/pldm/src/firmware_device.rs
```diff
@@ -0,0 +1,217 @@
...
+        loop {
+            // Phase 1: while in initiator mode, issue at most ONE outbound
+            // request per iteration. We deliberately fall through to the
+            // responder poll below (no `continue`) so an Update Agent command
+            // such as CancelUpdate is serviced between every RequestFirmwareData.
+            let initiator_active = self.cmd_interface.fd_ctx.should_start_initiator_mode();
+            if initiator_active {
+                let pldm_len = self
+                    .cmd_interface
+                    .generate_initiator_request(&mut fw_buf)
```
leongross, line 168: So what is the state when `generate_initiator_request` generates a buf len of `0`? Does that imply an error? This would then be handled with the `?` and we could get rid of the `pldm_len > 0`?

services/pldm/src/firmware_device.rs
```diff
@@ -0,0 +1,217 @@
...
+                        remote_eid,
+                        pldm_len,
+                        &mut fw_buf,
+                        requester_timeout_millis,
+                    )?;
+                    let resp_total_len =
+                        resp_len.checked_add(1).ok_or(PldmServiceError::Overflow)?;
+                    let resp = fw_buf
+                        .get_mut(..resp_total_len)
+                        .ok_or(PldmServiceError::Overflow)?;
```
leongross: Is `Overflow` an accurate error to use here?

services/pldm/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+//! │   FirmwareDevice<'a, O: FdOps, Cr, Cq>    │
+//! │     - cmd_interface: CmdInterface<'a, O>  │  PLDM FW-update state machine
+//! │     - responder_transport: inbound UA→FD  │
+//! │     - requester_transport: outbound FD→UA │
+//! └───────────┬──────────────┬────────────────┘
+//!             │              │ MctpPldmTransport<C: MctpClient>
+//!             ▼              ▼
+//! ┌──────────────────────────────────────────┐
+//! │   openprot-mctp-api                      │  Stack<C: MctpClient>
+//! └───────────┬───────────────────────────────┘
```
leongross, line 29: nit: improper formatting

services/pldm/src/lib.rs
```diff
@@ -0,0 +1,77 @@
...
+//! ## Architecture
+//!
+//! ```text
+//! ┌──────────────────────────┐
+//! │   Application / Firmware │  constructs FirmwareDevice, calls run_terminus()
+//! └───────────┬──────────────┘
+//!             │
+//!             ▼
+//! ┌──────────────────────────────────────────┐
+//! │   openprot-pldm-service                  │◄── this crate
```
leongross: nit: improper formatting

services/pldm/tests/base_host.rs
```diff
@@ -0,0 +1,445 @@
...
+impl Sender for BufferSender<'_> {
+    fn send_vectored(
+        &mut self,
+        mut fragmenter: Fragmenter,
+        payload: &[&[u8]],
+    ) -> mctp::Result<Tag> {
+        loop {
+            // Fragmenter requires the output buffer to be at least the payload
+            // MTU (255) plus the 4-byte MCTP transport header.
+            let mut buf = [0u8; 255 + 4];
```
leongross:
```suggestion
            let mut buf = [0u8; MTU + 4];
```

services/pldm/tests/base_host.rs
```diff
@@ -0,0 +1,445 @@
...
+            match fragmenter.fragment_vectored(payload, &mut buf) {
+                SendOutput::Packet(p) => self.packets.borrow_mut().push(p.to_vec()),
+                SendOutput::Complete { tag, .. } => return Ok(tag),
+                SendOutput::Error { err, .. } => return Err(err),
+            }
+        }
+    }
+
+    fn get_mtu(&self) -> usize {
+        255
```
leongross:
```suggestion
        MTU
```

services/pldm/tests/base_host.rs
```diff
@@ -0,0 +1,445 @@
...
+use pldm_common::message::firmware_update::get_status::ProgressPercent;
+use pldm_common::message::firmware_update::transfer_complete::TransferResult;
+use pldm_common::message::firmware_update::verify_complete::VerifyResult;
+use pldm_common::protocol::base::{PldmMsgType, PldmSupportedType, TransferOperationFlag};
+use pldm_common::protocol::firmware_update::{ComponentResponseCode, Descriptor};
+use pldm_common::util::fw_component::FirmwareComponent;
+use pldm_interface::firmware_device::fd_ops::{ComponentOperation, FdOps, FdOpsError};
+
+const FD_EID: u8 = 42;
+const UA_EID: u8 = 8;
```
leongross:
```suggestion
const UA_EID: u8 = 8;
const MTU: u8 = 255;
```

services/pldm/tests/base_host.rs
```diff
@@ -0,0 +1,445 @@
...
+
+struct FakeFdOps {
+    component_accepted: Cell<bool>,
+    download_bytes_received: Cell<usize>,
+    verified: Cell<bool>,
+    applied: Cell<bool>,
+    activated: Cell<bool>,
+}
+
+impl FdOps for FakeFdOps {
```
leongross: nit: just my opinion, use `Mock*` instead of `Fake*`

services/pldm/tests/firmware_update_host.rs
```diff
@@ -0,0 +1,563 @@
...
+            match fragmenter.fragment_vectored(payload, &mut buf) {
+                SendOutput::Packet(p) => self.packets.borrow_mut().push(p.to_vec()),
+                SendOutput::Complete { tag, .. } => return Ok(tag),
+                SendOutput::Error { err, .. } => return Err(err),
+            }
+        }
+    }
+
+    fn get_mtu(&self) -> usize {
+        255
```
leongross: see base_host.rs

services/pldm/tests/firmware_update_host.rs
leongross, line 1: It seems as if base_host and firmware_update_hosts share quite a lot of mock and setup code. Can this somehow be unified to reduce code duplication?

services/pldm/README.md
```diff
@@ -0,0 +1,127 @@
...
+process or platform-specific IPC bridge:
+
+```text
+┌──────────────────────────┐
+│   Application / Firmware │  constructs FirmwareDevice, calls run_terminus()
+└───────────┬──────────────┘
+            │
+            ▼
+┌──────────────────────────────────────────┐
+│   openprot-pldm-service                  │◄── this crate
```
leongross: nit: improper formatting

services/pldm/README.md
```diff
@@ -0,0 +1,127 @@
...
+│   FirmwareDevice<'a, O: FdOps, Cr, Cq>    │
+│     - cmd_interface: CmdInterface<'a, O>  │  PLDM FW-update state machine
+│     - responder_transport: inbound UA→FD  │
+│     - requester_transport: outbound FD→UA │
+└───────────┬──────────────┬────────────────┘
+            │              │ MctpPldmTransport<C: MctpClient>
+            ▼              ▼
+┌──────────────────────────────────────────┐
+│   openprot-mctp-api                      │  Stack<C: MctpClient>
+└───────────┬───────────────────────────────┘
```
leongross, line 35: nit: improper formatting

chrysh:
_A bit AI-y, but the problems it raises seem legit._

Two regression tests for `services/pldm/tests/base_host.rs` (they reuse the existing `MockFdOps` and `common.rs` wiring, so they drop straight in at the end of the file — no new imports needed).

**1. Responder must filter by UA EID.** Right now `run_terminus` acts on any inbound command regardless of source EID, so any endpoint on the bus can inject `SetTid`/`CancelUpdate`/etc. mid-update. This test currently **fails** (the FD applies a `SetTid` from EID 99); it passes once the responder path checks `meta.remote_eid` against the `remote_eid` passed to `run_terminus`.

```rust
/// Security regression: the FD must only act on commands from the Update
/// Agent EID it was told to serve (`run_terminus`'s `remote_eid`), not from
/// any endpoint that happens to be on the bus.
#[test]
fn responder_ignores_commands_from_unexpected_eid() {
    const ATTACKER_EID: u8 = 99;

    let fd_ops = MockFdOps {
        component_accepted: Cell::new(false),
        download_bytes_received: Cell::new(0),
        verified: Cell::new(false),
        applied: Cell::new(false),
        activated: Cell::new(false),
    };

    let ua_to_fd_packets = RefCell::new(Vec::new());
    let ua_server: RefCell<Server<_, 16>> = RefCell::new(Server::new(
        Eid(UA_EID),
        0,
        BufferSender { packets: &ua_to_fd_packets },
    ));

    let attacker_to_fd_packets = RefCell::new(Vec::new());
    let attacker_server: RefCell<Server<_, 16>> = RefCell::new(Server::new(
        Eid(ATTACKER_EID),
        0,
        BufferSender { packets: &attacker_to_fd_packets },
    ));

    let fd_to_ua_packets = RefCell::new(Vec::new());
    let fd_server: RefCell<Server<_, 16>> = RefCell::new(Server::new(
        Eid(FD_EID),
        0,
        BufferSender { packets: &fd_to_ua_packets },
    ));

    // Responder pump delivers BOTH queues, so whichever is pending is seen.
    let responder_client = DirectClientWithPump::new(&fd_server, || {
        transfer(&ua_to_fd_packets, &mut fd_server.borrow_mut());
        ua_to_fd_packets.borrow_mut().clear();
        transfer(&attacker_to_fd_packets, &mut fd_server.borrow_mut());
        attacker_to_fd_packets.borrow_mut().clear();
    });
    let responder_transport = MctpPldmTransport::new(responder_client);
    let requester_transport =
        MctpPldmTransport::new(DirectClientWithPump::new(&fd_server, || {}));

    let mut fd = FirmwareDevice::init(
        &fd_ops,
        &pldm_interface::config::PLDM_PROTOCOL_CAPABILITIES,
        responder_transport,
        requester_transport,
    );
    let mut fd_buf = [0u8; 1024];

    let mut run_fd_once =
        || match fd.run_terminus(UA_EID, &mut fd_buf, TIMEOUT_MILLIS, TIMEOUT_MILLIS) {
            Ok(()) => {}
            Err(PldmServiceError::Mctp(e)) if e.is_timeout() => {}
            Err(e) => panic!("firmware device failed: {e:?}"),
        };

    let mut buf = [0u8; 1024];
    // Attacker (EID 99) sends SetTid(0x99).
    let set_tid = SetTidRequest::new(0, PldmMsgType::Request, 0x99);
    let req_len = set_tid.encode(&mut buf).expect("encode attacker SetTid");
    let attacker_handle = attacker_server
        .borrow_mut()
        .req(FD_EID)
        .expect("attacker allocate request handle to FD");
    attacker_server
        .borrow_mut()
        .send(Some(attacker_handle), 0x01, None, None, false, &buf[..req_len])
        .expect("attacker send SetTid");
    run_fd_once();
    fd_to_ua_packets.borrow_mut().clear();

    // Legitimate UA (EID 8) queries the TID.
    let get_tid = GetTidRequest::new(1, PldmMsgType::Request);
    let req_len = get_tid.encode(&mut buf).expect("encode UA GetTid");
    let ua_handle = ua_server
        .borrow_mut()
        .req(FD_EID)
        .expect("UA allocate request handle to FD");
    ua_server
        .borrow_mut()
        .send(Some(ua_handle), 0x01, None, None, false, &buf[..req_len])
        .expect("UA send GetTid");
    run_fd_once();

    transfer(&fd_to_ua_packets, &mut ua_server.borrow_mut());
    fd_to_ua_packets.borrow_mut().clear();

    let mut resp = [0u8; 1024];
    let meta = ua_server
        .borrow_mut()
        .try_recv(ua_handle, &mut resp)
        .expect("GetTid response should be available");
    assert!(meta.payload_size >= 5, "GetTid response too short");
    assert_ne!(
        resp[4], 0x99,
        "FD acted on a SetTid from an unexpected EID ({ATTACKER_EID}); \
         the responder path must filter by the UA EID passed to run_terminus"
    );
}
```

**2. Pin the `run_terminus` idle-exit contract.** The doc says it "returns only on error", but an idle poll with a non-zero timeout exits with `Mctp(TimedOut)` — and both host tests rely on that as the "done" signal. This test makes that explicit so a future change that swallows idle timeouts (turning the loop into a hang) fails loudly here instead of hanging the suite. Passes as-is.

```rust
/// Contract regression: an idle `run_terminus` with a non-zero timeout must
/// exit with `Mctp(TimedOut)` rather than block or loop forever. Both host
/// tests use this as their "queue drained / done" signal.
#[test]
fn run_terminus_exits_with_timeout_when_idle() {
    let fd_ops = MockFdOps {
        component_accepted: Cell::new(false),
        download_bytes_received: Cell::new(0),
        verified: Cell::new(false),
        applied: Cell::new(false),
        activated: Cell::new(false),
    };

    let fd_to_ua_packets = RefCell::new(Vec::new());
    let fd_server: RefCell<Server<_, 16>> = RefCell::new(Server::new(
        Eid(FD_EID),
        0,
        BufferSender { packets: &fd_to_ua_packets },
    ));

    let responder_transport =
        MctpPldmTransport::new(DirectClientWithPump::new(&fd_server, || {}));
    let requester_transport =
        MctpPldmTransport::new(DirectClientWithPump::new(&fd_server, || {}));

    let mut fd = FirmwareDevice::init(
        &fd_ops,
        &pldm_interface::config::PLDM_PROTOCOL_CAPABILITIES,
        responder_transport,
        requester_transport,
    );
    let mut fd_buf = [0u8; 1024];

    match fd.run_terminus(UA_EID, &mut fd_buf, 5, 5) {
        Err(PldmServiceError::Mctp(e)) if e.is_timeout() => {}
        other => panic!(
            "idle run_terminus should exit with Mctp(TimedOut); got {other:?}"
        ),
    }
}
```

services/pldm/src/error.rs
```diff
@@ -0,0 +1,44 @@
...
+
+/// Errors returned by PLDM service operations.
+#[derive(Debug)]
+pub enum PldmServiceError {
+    /// An MCTP transport or stack error.
+    Mctp(MctpError),
+    /// A PLDM message handler error (codec failure, unsupported command, etc.).
+    MsgHandler(MsgHandlerError),
+    /// A buffer size or arithmetic overflow.
+    Overflow,
```
chrysh, line 19: Nice to haves:
- Make the variant more specific (BufferTooSmall or LengthOverflow) and carry context (requested vs available sizes) to aid diagnostics. Maybe convert Overflow into a struct with fields (needed, available) so callers/logs can report actionable info.

services/pldm/src/error.rs
```diff
@@ -0,0 +1,44 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            PldmServiceError::Mctp(e) => write!(f, "MCTP transport error: {e:?}"),
+            PldmServiceError::MsgHandler(e) => write!(f, "PLDM message handler error: {e:?}"),
+            PldmServiceError::Overflow => write!(f, "buffer size or arithmetic overflow"),
+        }
+    }
+}
+
+impl core::error::Error for PldmServiceError {}
```
chrysh, line 32: Nit:

The crate is no_std (lib.rs uses #![no_std]) but error.rs currently does impl core::error::Error for PldmServiceError; consider gating std::error::Error behind a "std" feature (#[cfg(feature = "std")] impl std::error::Error for PldmServiceError) so it compiles cleanly in strictly no_std builds.

services/pldm/src/firmware_device.rs
```diff
@@ -0,0 +1,223 @@
...
+        remote_eid: u8,
+        buf: &mut [u8],
+        timeout_millis: u32,
+        requester_timeout_millis: u32,
+    ) -> Result<(), PldmServiceError> {
+        let mut responder_listener = self
+            .responder_transport
+            .responder_listener(timeout_millis)?;
+        // Scratch buffer for FD-initiated (outbound) requests, reused across
+        // iterations rather than re-zeroed on every loop pass.
```
chrysh, line 156: Why is that safe, what if there is sensitive information in the buffer?

chrysh: Nit:

Will we ever have ressource contrainied embedded devices? In this case 1024 should go on the heap, or the caller should supply the buffer.

services/pldm/src/transport.rs
```diff
@@ -0,0 +1,264 @@
...
+        // Ensure the buffer is at least large enough to hold the framed
+        // request before handing it off.
+        let req_total_len = payload_size
+            .checked_add(1)
+            .filter(|&total_len| total_len <= buf.len())
+            .ok_or(PldmServiceError::Overflow)?;
+
+        // Invoke the handler to process the request in-place.  The handler is
+        // given the whole buffer so the response may exceed the request size.
+        let resp_total_len = handler(buf, req_total_len, meta.remote_eid)?;
```
chrysh, line 247: Defensive, not a known exploit:

`respond_once` hands the handler the whole `buf`, and the `run_terminus` closure ignores `req_total_len`, so `handle_responder_msg` ends up decoding from `buf[1..]` — i.e. past the bytes actually received this iteration, into whatever a previous message left behind.

```rust
handler(&mut buf[..req_total_len], req_total_len, meta.remote_eid)
```

Cheap, and it means decoding can never reach beyond the current message regardless of what pldm-lib does internally. Removes the foot-gun rather than relying on every command decoder to bound itself.

services/pldm/src/firmware_device.rs
chrysh: API-improvement: change the signature to return Result<Option<usize>, MsgHandlerError> (or Result<usize, MsgHandlerError> where usize>0 always indicates a request and None/Err for no request/error). Then caller becomes: match self.cmd_interface.generate_initiator_request(&mut fw_buf).map_err(...)? { Some(pldm_len) => { /* send */ } None => {} } This is clearer and avoids encoding a special-case 0.

chrysh:
Two more regression guards for the EID fix. These aren't new findings — they lock in that the filter covers paths the current `unexpected_eid_host.rs` (SetTid) doesn't: the **firmware-update state machine** (#1) and **not leaking a response at all** (#2). I checked both fail if the `source_eid != remote_eid` guard is removed (state advances to `1`; a response leaks), so they're not vacuous.

I also collapsed the per-test MCTP wiring into one `with_bus` scope helper, so each test reads as intent rather than plumbing — a third case costs ~5 lines. Drop these into `unexpected_eid_host.rs` (reuse its `MockFdOps`), or as a new `tests/eid_compact.rs` with a `rust_test` target like the others (boilerplate collapsed below).

```rust
/// Stand up UA (EID 8), a rogue endpoint (EID 99) and the FD (EID 42) over
/// in-memory MCTP, then run `body`. `body` gets a `send(src_eid, pldm_bytes)`
/// closure that delivers one command, runs the terminus once, and returns the
/// response the *sender* receives — `None` if the FD stayed silent.
fn with_bus(body: impl FnOnce(&MockFdOps, &mut dyn FnMut(u8, &[u8]) -> Option<Vec<u8>>)) {
    let fd_ops = MockFdOps::default();
    let (ua_q, atk_q, fd_q) = (
        RefCell::new(Vec::new()),
        RefCell::new(Vec::new()),
        RefCell::new(Vec::new()),
    );
    let ua = RefCell::new(Server::<_, 16>::new(Eid(UA_EID), 0, BufferSender { packets: &ua_q }));
    let atk =
        RefCell::new(Server::<_, 16>::new(Eid(ATTACKER_EID), 0, BufferSender { packets: &atk_q }));
    let fd_srv = RefCell::new(Server::<_, 16>::new(Eid(FD_EID), 0, BufferSender { packets: &fd_q }));

    // One responder pump drains both senders' queues into the FD before recv.
    let responder = DirectClientWithPump::new(&fd_srv, || {
        transfer(&ua_q, &mut fd_srv.borrow_mut());
        ua_q.borrow_mut().clear();
        transfer(&atk_q, &mut fd_srv.borrow_mut());
        atk_q.borrow_mut().clear();
    });
    let requester = DirectClientWithPump::new(&fd_srv, || {});
    let mut fd = FirmwareDevice::init(
        &fd_ops,
        &pldm_interface::config::PLDM_PROTOCOL_CAPABILITIES,
        MctpPldmTransport::new(responder),
        MctpPldmTransport::new(requester),
    );
    let mut fd_buf = [0u8; 1024];

    let mut send = |src_eid: u8, payload: &[u8]| -> Option<Vec<u8>> {
        let sender = if src_eid == UA_EID { &ua } else { &atk };
        let h = sender.borrow_mut().req(FD_EID).expect("req handle");
        sender
            .borrow_mut()
            .send(Some(h), 0x01, None, None, false, payload)
            .expect("send");
        let _ = fd.run_terminus(UA_EID, &mut fd_buf, TIMEOUT_MILLIS, TIMEOUT_MILLIS);
        transfer(&fd_q, &mut sender.borrow_mut());
        fd_q.borrow_mut().clear();
        let mut resp = [0u8; 1024];
        sender
            .borrow_mut()
            .try_recv(h, &mut resp)
            .map(|m| resp[..m.payload_size].to_vec())
    };

    body(&fd_ops, &mut send);
}

/// #1 — a rogue endpoint must not drive the firmware-update state machine.
#[test]
fn rogue_eid_cannot_drive_update() {
    with_bus(|fd_ops, send| {
        send(ATTACKER_EID, &request_update()); // dropped, no response expected
        let status = send(UA_EID, &get_status()).expect("UA gets a status reply");
        let st = GetStatusResponse::decode(&status).expect("decode status");
        assert_eq!(
            st.current_state,
            FirmwareDeviceState::Idle as u8,
            "rogue RequestUpdate advanced the FD state machine"
        );
        assert!(!fd_ops.component_accepted.get());
    });
}

/// #2 (bonus) — the FD must not disclose state to a rogue endpoint (no response).
#[test]
fn rogue_eid_gets_no_response() {
    with_bus(|_fd_ops, send| {
        assert!(
            send(ATTACKER_EID, &get_status()).is_none(),
            "FD replied to a GetStatus from an unexpected EID"
        );
    });
}
```

<details>
<summary>Encode helpers + minimal <code>MockFdOps</code> + imports (boilerplate)</summary>

```rust
use core::cell::{Cell, RefCell};

use mctp::Eid;
use openprot_mctp_server::Server;
use openprot_pldm_service::firmware_device::FirmwareDevice;
use openprot_pldm_service::MctpPldmTransport;
use pldm_common::codec::PldmCodec;
use pldm_common::message::firmware_update::apply_complete::ApplyResult;
use pldm_common::message::firmware_update::get_fw_params::FirmwareParameters;
use pldm_common::message::firmware_update::get_status::{
    GetStatusRequest, GetStatusResponse, ProgressPercent,
};
use pldm_common::message::firmware_update::request_update::RequestUpdateRequest;
use pldm_common::message::firmware_update::transfer_complete::TransferResult;
use pldm_common::message::firmware_update::verify_complete::VerifyResult;
use pldm_common::protocol::base::PldmMsgType;
use pldm_common::protocol::firmware_update::{
    ComponentResponseCode, Descriptor, FirmwareDeviceState, PldmFirmwareString, VersionStringType,
    PLDM_FWUP_IMAGE_SET_VER_STR_MAX_LEN,
};
use pldm_common::util::fw_component::FirmwareComponent;
use pldm_interface::firmware_device::fd_ops::{ComponentOperation, FdOps, FdOpsError};

mod common;
use common::{transfer, BufferSender, DirectClientWithPump, FD_EID, TIMEOUT_MILLIS, UA_EID};

const ATTACKER_EID: u8 = 99;

fn request_update() -> Vec<u8> {
    let ver = PldmFirmwareString {
        str_type: VersionStringType::Ascii as u8,
        str_len: 4,
        str_data: {
            let mut d = [0u8; PLDM_FWUP_IMAGE_SET_VER_STR_MAX_LEN];
            d[..4].copy_from_slice(b"v1.0");
            d
        },
    };
    let mut b = [0u8; 128];
    let n = RequestUpdateRequest::new(0, PldmMsgType::Request, 1024, 1, 1, 0, &ver)
        .encode(&mut b)
        .unwrap();
    b[..n].to_vec()
}

fn get_status() -> Vec<u8> {
    let mut b = [0u8; 64];
    let n = GetStatusRequest::new(0, PldmMsgType::Request).encode(&mut b).unwrap();
    b[..n].to_vec()
}

// Only `component_accepted` is observed here; the rest are trait stubs.
#[derive(Default)]
struct MockFdOps {
    component_accepted: Cell<bool>,
}

impl FdOps for MockFdOps {
    fn get_device_identifiers(&self, _d: &mut [Descriptor]) -> Result<usize, FdOpsError> { Ok(0) }
    fn get_firmware_parms(&self, p: &mut FirmwareParameters) -> Result<(), FdOpsError> {
        *p = FirmwareParameters::default();
        Ok(())
    }
    fn get_xfer_size(&self, s: usize) -> Result<usize, FdOpsError> { Ok(s.min(512)) }
    fn handle_component(
        &self,
        _c: &FirmwareComponent,
        _p: &FirmwareParameters,
        _o: ComponentOperation,
    ) -> Result<ComponentResponseCode, FdOpsError> {
        self.component_accepted.set(true);
        Ok(ComponentResponseCode::CompCanBeUpdated)
    }
    fn query_download_offset_and_length(
        &self,
        _c: &FirmwareComponent,
    ) -> Result<(usize, usize), FdOpsError> { Ok((0, 1024)) }
    fn download_fw_data(
        &self,
        _o: usize,
        _d: &[u8],
        _c: &FirmwareComponent,
    ) -> Result<TransferResult, FdOpsError> { Ok(TransferResult::TransferSuccess) }
    fn is_download_complete(&self, _c: &FirmwareComponent) -> bool { true }
    fn query_download_progress(
        &self,
        _c: &FirmwareComponent,
        _p: &mut ProgressPercent,
    ) -> Result<(), FdOpsError> { Ok(()) }
    fn verify(
        &self,
        _c: &FirmwareComponent,
        _p: &mut ProgressPercent,
    ) -> Result<VerifyResult, FdOpsError> { Ok(VerifyResult::VerifySuccess) }
    fn apply(
        &self,
        _c: &FirmwareComponent,
        _p: &mut ProgressPercent,
    ) -> Result<ApplyResult, FdOpsError> { Ok(ApplyResult::ApplySuccess) }
    fn activate(&self, _s: u8, _e: &mut u16) -> Result<u8, FdOpsError> { Ok(0) }
    fn cancel_update_component(&self, _c: &FirmwareComponent) -> Result<(), FdOpsError> { Ok(()) }
}
```
</details>

services/pldm/tests/base_host.rs
```diff
@@ -0,0 +1,343 @@
...
+            None,
+            None,
+            false,
+            &ua_req_buf[1..req_len],
+        )
+        .expect("send request_update payload");
+
+    // Runs `FirmwareDevice::run_terminus` until its inbound queue is drained.
+    // `run_terminus` loops until its responder listener has nothing left, at
+    // which point it returns Mctp(TimedOut); that terminating timeout means
```
chrysh, line 207: API-change suggestion (if you want code-level clarity rather than comment):
Change run_terminus signature to return a richer result, e.g.:
```rust
pub enum RunTerminusResult { Completed, StoppedByError(PldmServiceError) }
pub fn run_terminus(...) -> Result<RunTerminusResult, PldmServiceError>
```
Then when the listener times out return `Ok(RunTerminusResult::Completed) `instead of `Err(Mctp(TimedOut))`. Callers no longer need to special-case a timeout error.

chrysh (changes requested):

services/pldm/README.md
```diff
@@ -0,0 +1,127 @@
...
+`recv_and_respond`, `respond_once`) and `FirmwareDevice::run_terminus` use the
+same flat-buffer convention:
+
+```text
+buf[0]   : MCTP message-type byte (0x01) — managed by MctpPldmTransport
+buf[1..] : PLDM request / response bytes
+```
+
+Size the buffer to accommodate the largest PLDM message your application
+expects (typically ≤ 4096 bytes; smaller for embedded targets). See
```
chrysh, line 96: Nit:

This says "typically ≤ 4096 bytes", but `FD_IPC_MAX_MSG` in `firmware_device.rs` is 1024 and is the hard cap on FD-initiated requests (`fw_buf = [0u8; FD_IPC_MAX_MSG]`). A reader budgeting 4096 would assume 4× the headroom the code allows. Suggest quoting the actual 1024 (or deriving this number from the constant).

services/pldm/src/firmware_device.rs
```diff
@@ -0,0 +1,223 @@
...
+use pldm_interface::cmd_interface::CmdInterface;
+use pldm_interface::control_context::ProtocolCapability;
+use pldm_interface::firmware_device::fd_context::FirmwareDeviceContext;
+use pldm_interface::firmware_device::fd_ops::FdOps;
+
+use crate::error::PldmServiceError;
+use crate::transport::MctpPldmTransport;
+
+/// Maximum PLDM-over-MCTP message size (MCTP-type byte + PLDM payload).
+pub const FD_IPC_MAX_MSG: usize = 1024;
```
chrysh: The `_IPC_` in `FD_IPC_MAX_MSG` refers to the IPC bridge this PR removes — `FirmwareDevice` now owns both transports directly. The name points at an architecture that no longer exists; `FD_MAX_MSG` would match what it is.

CourtneyDrant:
Thanks,
The handler used by respond_once in run_terminus leaves before reaching handle_responder_msg when it doesn't recognize the eid as being from our designated EID.  How do we currently impact the FD state machine without executing the handle_responder_msg?  Is that possible?

chrysh:
> Thanks, The handler used by respond_once in run_terminus leaves before reaching handle_responder_msg when it doesn't recognize the eid as being from our designated EID. How do we currently impact the FD state machine without executing the handle_responder_msg? Is that possible?

@CourtneyDrant Not currently, it is meant a regression guard for future code changes.

CourtneyDrant:
> > Thanks, The handler used by respond_once in run_terminus leaves before reaching handle_responder_msg when it doesn't recognize the eid as being from our designated EID. How do we currently impact the FD state machine without executing the handle_responder_msg? Is that possible?
> 
> @CourtneyDrant Not currently, it is meant a regression guard for future code changes.
Okay, so your concern is having a regression test that verifies attacker EIDs can't influence the FD state machine.  So essentially, we need an attacker Type 5 regression test.

services/pldm/src/firmware_device.rs
```diff
@@ -0,0 +1,250 @@
...
+                {
+                    Some(pldm_len) => {
+                        let resp_len = self.requester_transport.send_request(
+                            remote_eid,
+                            pldm_len,
+                            &mut fw_buf,
+                            requester_timeout_millis,
+                        )?;
+                        let resp_total_len =
+                            // resp_len.checked_add(1).ok_or(PldmServiceError::Overflow)?;
```
chrysh: Leftover commented-out line — please remove.

services/pldm/src/transport.rs
```diff
@@ -0,0 +1,264 @@
...
+    /// [`FirmwareDevice::run_terminus`](crate::firmware_device::FirmwareDevice::run_terminus),
+    /// which interleaves FD-initiated requests) should use
+    /// [`responder_listener`](Self::responder_listener) and
+    /// [`respond_once`](Self::respond_once) instead, to avoid a window with no
+    /// listener registered in which an inbound request could be dropped by
+    /// the underlying MCTP stack.
+    ///
+    /// # Errors
+    ///
+    /// Returns [`PldmServiceError::Overflow`] if `buf` is too small.
```
chrysh: This still references `PldmServiceError::Overflow`, which no longer exists — `respond_once`'s Errors section was updated, this one was missed.

chrysh (changes requested):
Thanks — most of the last round is addressed and the interface reads much cleaner now.

services/pldm/src/firmware_device.rs
chrysh: The rename made it into the code, but README.md line 97 still says FD_IPC_MAX_MSG.

services/pldm/tests/base_host.rs
chrysh: Thanks for adding the enum. As-is, though, Completed is documented as unreachable and the idle listener timeout still surfaces as StoppedByError(Mctp(TimedOut)), so this test still special-cases is_timeout(). If run_terminus_inner maps the idle-timeout to Completed, callers can drop that special case — which was the point of the enum.

CourtneyDrant:
Can we move any new regression requests to issues and finish this PR?

chrysh (approved):
Yes, let's finish this PR. The RunTerminusResult idle-timeout mapping (timeout -> Completed so callers can drop the is_timeout() special case) can move to a follow-up issue — feel free to open it, or I will.

One small thing to fix here: README.md line 97 still references FD_IPC_MAX_MSG; the constant is now FD_MAX_MSG.

Approving already so this isn't blocked on me.

chrysh:
Filed the timeout-mapping follow-up as #429.

leongross (approved):
Resolve the merge conflict and then we can get this merged.

wmaroneAMD:
The merge conflicts are in MODULE.bazel.lock - can we deploy the Bazel conflict autoresolver?

https://bazel.build/external/lockfile#automatic-resolution

leongross:
> The merge conflicts are in MODULE.bazel.lock - can we deploy the Bazel conflict autoresolver?
> 
> [bazel.build/external/lockfile#automatic-resolution](https://bazel.build/external/lockfile#automatic-resolution)

That looks very promising, good catch!

PR #380, opened 2026-07-29
GPIO interrupt test

lxuxx:
1. Open drain mode operations are updated.
2. GPIO interrupts are tested with GPIOI loopback connection.

### DIFF
--- a/target/ast10x0/tests/peripherals/gpio/gpioi_loopback/target.rs
+++ b/target/ast10x0/tests/peripherals/gpio/gpioi_loopback/target.rs
@@ -1,0 +1,196 @@
+// Licensed under the Apache-2.0 license
+// SPDX-License-Identifier: Apache-2.0
+
+//! AST10x0 GPIOI loopback test target.
+
+#![no_std]
+#![no_main]
+
+use ast10x0_peripherals::gpio::{gpioi, ActiveLow, GpioExt, InterruptMode};
+use ast10x0_peripherals::scu::{pinctrl, ScuRegisters};
+use console_backend::console_backend_write_all;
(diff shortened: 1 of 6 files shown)


JesseMelon:
This looks good.

EdgeBoth leaves stale sensitivity bits
Every other mode writes all three int_sen_t0/t1/t2 registers, but EdgeBoth only sets t2 and never clears t0/t1. When a pin is reconfigured from a level/edge mode to EdgeBoth, t0/t1 stay dirty. This happens during the test, and presumably it still passes Is this intended?

The now dead test_gpio_output() can be removed.

As soon I have someone in the office to attach the jumper I can verify the test passes. Otherwise, I will approve after the above!

lxuxx:
Yes, both-edge only care about type2 as shown in the table.
image

JesseMelon (approved):

target/ast10x0/tests/peripherals/gpio/gpioi_loopback/target.rs
```diff
@@ -0,0 +1,165 @@
...
+use target_common::{declare_target, TargetInterface};
+use {console_backend as _, entry as _};
+
+pub struct Target {}
+
+const GPIO_LEVEL_DELAY_CYCLES: u32 = 200_000_000;
+
+fn test_gpio_loopback() -> bool {
+    // The loopback requires a jumper between the fixture board's exposed
+    // GPIOI0/I2C2 SCL pin and GPIOI1/I2C2 SDA pin.
```
leongross, line 22: It would be helpful to have this in the BOARDS.md setup file for the AST1060 with a picture highlighting the referenced GPIO pins for easy reproduction.
@JesseMelon @rusty1968

PR #386, opened 2026-07-30
Attestation libraries

gmandyam:
Add attestation service.  Initial cut.


third_party/crates_io/Cargo.toml
```diff
@@ -40,7 +40,7 @@ zerocopy = { version = "0.8.48", default-features = false, features = ["derive"]
 zeroize = { version = "1.8", default-features = false, features = ["derive"] }
 
 cortex-m = { version = "0.7.7", features = ["critical-section-single-core"] }
-cortex-m-rt = "0.7.5"
+cortex-m-rt = "=0.7.5"
```
rusty1968: We don't need the exact = pin here. The committed Cargo.lock already resolves cortex-m-rt (and cortex-m-rt-macros) to exactly 0.7.5 with a checksum, so every build that respects the lockfile is already reproducible — the = on the requirement doesn't add any determinism on top of that.

services/attest/api/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+# Licensed under the Apache-2.0 license
+# SPDX-License-Identifier: Apache-2.0
```
rusty1968: Can we drop Cargo.toml? This is a Bazel project and the crate is already fully defined in BUILD.bazel — targets and deps included.

services/attest/producer/Cargo.toml
```diff
@@ -0,0 +1,22 @@
+# Licensed under the Apache-2.0 license
```
rusty1968: Can we drop Cargo.toml? This is a Bazel project and the crate is already fully defined in BUILD.bazel — targets and deps included.

rusty1968:
Quick note on Cargo.toml files: this is a Bazel project, so the only manifest we actually need is Cargo.toml — that's what drives crate_universe / external crate resolution for Bazel. All the per-crate Cargo.toml files scattered under services (attest, spdm, mctp, etc.) aren't used to build or test anything; the real target and dependency definitions live in each crate's BUILD.bazel.

services/attest/api/src/types.rs
```diff
@@ -0,0 +1,66 @@
...
+    pub hw_version: String,
+    pub cert_cache_ttl: Duration,
+}
+
+/// Hardware-backed signing operations.
+///
+/// Production: implemented by a Caliptra mailbox driver.
+/// Testing: implement with a software key (`SoftwareAttestProducer` in the
+/// producer crate behind `test-support`).
+pub trait CaliptraSigner: Send + Sync {
```
rusty1968: Small naming nit: CaliptraSigner lives in the "platform-independent" API crate, but the name (and sign_es384/alias_cert_der) bakes in a specific vendor and algorithm.

services/attest/producer/src/dice_identity.rs
```diff
@@ -0,0 +1,76 @@
...
+    if chain.len() < 2 {
+        return Err(AttestError::Caliptra(
+            "DICE certificate chain must have at least two certificates (leaf + one CA)".into(),
+        ));
+    }
+    Ok(CertChain(chain))
+}
+
+/// Return just the DER-encoded Alias (leaf) certificate.
+pub fn alias_cert(signer: &dyn CaliptraSigner) -> Result<Vec<u8>, AttestError> {
```
rusty1968: alias_cert is dead — I grepped the whole tree and its only references are the definition itself and the one unit test that exists solely to test it. This is just a pass-thru method invoked only by the test.

rusty1968:
As written, both attest crates are std-based and heap-allocating. Please make sure all crates are no_std.

services/attest/producer/src/builder.rs
```diff
@@ -0,0 +1,280 @@
...
+//!     << { 1: -35, 33: [cert-chain...] } >>,   ; protected header
+//!     {},                                        ; unprotected header
+//!     << 61( { ...CWT claims... } ) >>,          ; payload (EAT CWT)
+//!     h'...'                                     ; ES384 signature
+//!   ]
+//! )
+//! ```
+//!
+//! Claim key numbers follow RFC 9711 and the OCP-EAT profile.
+
```
rusty1968, line 20:
```suggestion
#![no_std]
#![forbid(unsafe_code)]
```

rusty1968 (changes requested):

services/attest/producer/src/builder.rs
```diff
@@ -0,0 +1,280 @@
...
+    let chain = signer.cert_chain_der()?;
+
+    let mut claims: Vec<(Value, Value)> = Vec::new();
+
+    claims.push((
+        Value::Integer(CLAIM_ISS.into()),
+        Value::Text("https://openprot.example/caliptra/device".into()),
+    ));
+
+    let iat = std::time::SystemTime::now()
```
rusty1968: The crate must be `no_std.` You are using O/S facilities in a crate that is going to be consumed in an embedded environment.

rusty1968:
One thing I noticed: dice_identity.rs and measurements.rs are each basically a whole module wrapping a single pub fn that's only ~1-3 lines — collect is a loop + extend, cert_chain is one call + a length check, and alias_cert is a straight pass-through (and dead). It's a bit of a "one file per function" thing, where the module boundary costs more than the logic it's hiding. These would be better as private helpers/methods behind HwAttestProducer.

third_party/crates_io/Cargo.toml
gmandyam: Changed accordingly.  Exact pin removed.

services/attest/api/Cargo.toml
gmandyam: Dropped.

services/attest/api/src/types.rs
gmandyam: Renamed to HWSigner

services/attest/producer/Cargo.toml
gmandyam: Dropped.

services/attest/producer/src/dice_identity.rs
gmandyam: Removed function and unit test

rusty1968:
## openprot-secure-coding — violation

`zeroize` is added as a dependency (`producer/Cargo.toml:19`, `producer/BUILD.bazel:16`) and
the README (`producer/README.md:81`) documents it as "Zero-on-drop for intermediate key
material and sensitive buffers" — but it's never imported or called anywhere in the source.
Dead dependency, and the security claim in the README is currently false.

gmandyam:
> As written, both attest crates are std-based and heap-allocating. Please make sure all crates are no_std.

Moved to heapless.

services/attest/producer/src/builder.rs
gmandyam: stdlib dependencies are removed.

services/attest/producer/src/builder.rs
gmandyam: Added.  See lib.rs files.

gmandyam:
> ## openprot-secure-coding — violation
> `zeroize` is added as a dependency (`producer/Cargo.toml:19`, `producer/BUILD.bazel:16`) and the README (`producer/README.md:81`) documents it as "Zero-on-drop for intermediate key material and sensitive buffers" — but it's never imported or called anywhere in the source. Dead dependency, and the security claim in the README is currently false.

Removed and README updated.

services/attest/producer/src/lib.rs
```diff
@@ -0,0 +1,21 @@
...
+#![forbid(unsafe_code)]
+
+pub mod builder;
+pub mod dice_identity;
+pub mod measurements;
+mod signer;
+
+pub use signer::HwAttestProducer;
+
+#[cfg(feature = "test-support")]
```
rusty1968, line 21: Can this live under mod tests?

rusty1968 (changes requested):
LGTM: [x]

## Correctness

- `producer/src/builder.rs` `build()`: the `ueid` claim (key 256) is hard-coded to a constant placeholder (`ueid[0] = 0x01`, the remaining 28 bytes always zero) for *every* device, in the shared `build()` path used by both `HwAttestProducer` (production) and `SoftwareAttestProducer`. The README documents this claim as "Unique Entity Identifier from Caliptra CDI", but nothing derives it from device identity — every token from every device will carry the same UEID. This defeats the purpose of the claim and should either be wired to a real per-device source (e.g. via `HwSigner`/DICE identity) or clearly gated so it can't ship silently wrong (a `// TODO` alone is not enough given this is the production code path, not just the test stub).
- `third_party/crates_io/Cargo.lock`: diffed against current `main` tip (not just this PR's stale base), this PR's lockfile reintroduces `pldm-common`/`pldm-interface` (already removed on `main`), is missing `statig` (added on `main` since this branch's last lockfile rebuild), and downgrades ~15+ unrelated crates (`bitflags`, `bytes`, `clap`/`clap_builder`, `hashbrown`, `foldhash`, `log`, `memchr`, `smallvec`, `uuid`, `ruzstd`, `syn`, `zerocopy`, …) that have nothing to do with `ciborium`/`minicbor`. This needs a lockfile refresh/rebase before merge — merging as-is risks silently resurrecting removed dependencies and regressing pinned versions repo-wide.
- Both `services/attest/api/README.md` and `services/attest/producer/README.md` document an API that no longer matches the code: the README shows a `CaliptraSigner` trait with `sign_es384`/`alias_cert_der`/`cert_chain_der(&self) -> Result<Vec<Vec<u8>>, _>`, and an `AttestProducer::generate_token(&self, ...) -> Result<Vec<u8>, _>` / `cert_chain(&self) -> Result<CertChain, _>`. The actual code (`traits.rs`, `types.rs`) has been reworked to `no_std`/heapless out-parameter signatures (`HwSigner`, `generate_token(..., iat, out: &mut Vec<u8, MAX_TOKEN_SIZE>)`, `cert_chain(&self, buf: &mut Vec<...>)`), and the producer README's usage examples (`Arc::new(caliptra_driver)`, `Box::new(UefiFirmwareMeasurements)`, `let token: Vec<u8> = ...generate_token(...)`) don't compile against the real, reference-based, no-alloc constructors. These docs will actively mislead the first integrator who copies them. The README's claim table is also missing the `sw-name`(14)/`sw-version`(15) claims that `builder.rs` actually emits.

## No dynamic allocation (openprot-no-alloc)

No violations found. Verified across the full diff: both crates declare `#![no_std]` (`api/src/lib.rs`, `producer/src/lib.rs`), and every buffer in `consts.rs`, `error.rs`, `traits.rs`, `types.rs`, `builder.rs`, `dice_identity.rs`, `measurements.rs`, and `signer.rs` uses `heapless::{Vec, String}` with named fixed capacities from `consts.rs`. The only place `std`-only crates (`ciborium`) appear is `producer/tests/producer_integration.rs`, a host-side integration-test binary, which is the documented exemption.

## Panic-free / explicit errors (openprot-no-panic)

- `producer/src/measurements.rs::test_caliptra_measurements()` is gated by `#[cfg(feature = "test-support")]`, **not** `#[cfg(test)]` — it is reachable at runtime from `SoftwareAttestProducer::generate_token`, a build configuration the README documents as legitimate ("software stub enabled, no Caliptra hardware required"). It contains 4 `.unwrap()` calls (`push_str`, `extend_from_slice`, `push`). These are practically infallible given the hardcoded literal lengths, but the rule is unconditional outside `#[cfg(test)]`. Suggested fix (included in the diff): return `Result<Vec<Measurement, MAX_MEASUREMENTS>, AttestError>` and propagate with `?`/`map_err`, matching the style already used everywhere else in this crate; update the one call site in `signer.rs`.
- Everywhere else (`builder.rs` production path, `signer.rs` non-test-support impls, `dice_identity.rs` production fn) consistently uses `map_err(...)?` — good, this is the only outlier.

## Secrets, crypto, and register access (openprot-secure-coding)

- The prior finding (unused `zeroize` dependency) is resolved: `zeroize` no longer appears anywhere, and there is in fact no local sensitive/secret buffer in this diff to zeroize — the private Alias Key never leaves the (out-of-scope) Caliptra mailbox driver, and `HwSigner::sign()` only returns a public signature. Nothing to flag here.
- No `unsafe` anywhere (`#![forbid(unsafe_code)]` in both crate roots), no secret-byte comparisons, and no key material appears in `Debug`/error output.
- `builder::build()` stack-allocates two independent `MAX_TOKEN_SIZE` (8 KiB each) scratch buffers (`payload_scratch`, `cose_scratch`) simultaneously, plus a 128-byte header scratch and the caller-owned `out` buffer. Worth confirming this fits the target RoT's stack budget (~16-24 KiB of scratch alone) — not blocking, but should be measured/documented for an embedded target.

## Verbosity / over-engineering (ponytail-review)

- `builder.rs::build()` has ~30 near-identical `e.<method>(...).map_err(|_| AttestError::Cbor)?;` lines. The previously-flagged shape (`claims.push((Value::Integer(...), Value::...))` → a `claim()` helper) doesn't map cleanly onto the new streaming `minicbor::Encoder` API, but the repeated `.map_err(|_| AttestError::Cbor)?` boilerplate could still be collapsed with a small local macro or closure. Nit, not blocking.
- The stub DER-cert bytes (`[0x30, 0x00]`) are constructed inline and duplicated across `signer.rs` (`StubSigner`, `SoftwareAttestProducer::cert_chain`), `dice_identity.rs` tests, and `builder.rs` tests — a shared `fn stub_cert() -> [u8; 2]` test helper would remove the duplication. Low priority, test-only code.
- `decode_outer`/`find_claim*` helpers in `builder.rs` tests (via `minicbor`) and `producer_integration.rs` (via `ciborium`) look like duplicated test scaffolding, but they intentionally use two different CBOR libraries to cross-validate the encoder against an independent decoder — that's reasonable, not bloat. The comment in `builder.rs`'s `decode_outer` ("Use ciborium only in tests") is simply stale/wrong now (the code uses `minicbor::Decoder`); fixed in the diff.

## Readability & Maintainability

- `HwAttestProducer.providers: Vec<&'a dyn MeasurementProvider, 8>` uses a bare magic-number capacity where every other buffer in this crate uses a named `MAX_*` constant; fixed in the diff (`MAX_PROVIDERS`).
- `third_party/crates_io/Cargo.toml`: `ciborium`/`minicbor` were inserted directly under the `# Host tool crates needed by pigweed pw_kernel/tooling` comment even though `minicbor` is a production, `no_std` dependency of the producer crate (not host tooling). Fixed with its own comment block in the diff.

## Commit Message

- This PR's commit history mixes Conventional-Commit-style prefixes (`fix:`, `deps:`, `build:`, `style:`) that don't match the plainer, component-first style used elsewhere in this repo's history (e.g. `orchestrator: ...`, `Add X`, `Fix Y`). Not blocking if this gets squash-merged, but worth a single descriptive commit message on merge rather than carrying the 20-commit fixup trail (several of which are literally "fix: resolve clippy warnings" / "fix: remove unused import" against this same PR's own earlier commits).

chrysh:
@rusty1968 @gmandyam Are we adding AI skills to the repo now?

services/attest/producer/src/builder.rs
```diff
@@ -0,0 +1,428 @@
...
+    }
+}
+
+/// Build and sign a complete OCP-EAT token into `out`.
+///
+/// `iat` is a Unix timestamp (seconds since epoch) supplied by the caller.
+pub(crate) fn build(
+    config: &AttestConfig,
+    signer: &dyn HwSigner,
+    measurements: &Vec<Measurement, MAX_MEASUREMENTS>,
```
chrysh: Take a slice here. `build()` only calls `.len()` and iterates, so the capacity in
the type buys nothing and makes `MAX_MEASUREMENTS` part of the signature: a
caller holding a `Vec<Measurement, 8>` or a `[Measurement; 3]` has to copy into a
`Vec<_, 16>` first, and bumping the constant becomes a breaking change.
`signer.rs:58` already passes `&meas`, which coerces on its own, so no call site
changes.

```suggestion
    measurements: &[Measurement],
```

services/attest/producer/src/builder.rs
```diff
@@ -0,0 +1,428 @@
...
+//! )
+//! ```
+//!
+//! Claim key numbers follow RFC 9711 and the OCP-EAT profile.
+
+use heapless::Vec;
+use minicbor::encode::Write as CborWrite;
+use minicbor::Encoder;
+
+use openprot_attest_api::consts::{MAX_CERT_SIZE, MAX_CHAIN_LEN, MAX_MEASUREMENTS, MAX_TOKEN_SIZE};
```
chrysh: Apply together with the line-82 change, not on its own: once `build()` takes a
slice, `MAX_MEASUREMENTS` is unused outside the test module, which has its own
import at line 236. Applied alone this breaks the build.

```suggestion
use openprot_attest_api::consts::{MAX_CERT_SIZE, MAX_CHAIN_LEN, MAX_TOKEN_SIZE};
```

services/attest/producer/src/builder.rs
```diff
@@ -0,0 +1,428 @@
...
+    pos: usize,
+}
+
+impl<'a> BufWriter<'a> {
+    fn new(buf: &'a mut [u8]) -> Self {
+        Self { buf, pos: 0 }
+    }
+}
+
+impl<'a> CborWrite for BufWriter<'a> {
```
chrysh: The lifetime is never referenced in the impl body, so `'_` says the same thing.
(Moot if you take the `Cursor` suggestion on line 51.)

```suggestion
impl CborWrite for BufWriter<'_> {
```

services/attest/producer/src/builder.rs
```diff
@@ -0,0 +1,428 @@
...
+    type Error = AttestError;
+    fn write_all(&mut self, data: &[u8]) -> Result<(), AttestError> {
+        let end = self.pos + data.len();
```
chrysh: `self.pos + data.len()` overflows silently in release and panics in debug, in a
crate that forbids panics. The crate already has the right error for it.
(Moot if you take the `Cursor` suggestion on line 51.)

```suggestion
        let end = self.pos.checked_add(data.len()).ok_or(AttestError::BufferFull)?;
```

services/attest/producer/src/builder.rs
```diff
@@ -0,0 +1,428 @@
...
+        if end > self.buf.len() {
+            return Err(AttestError::BufferFull);
+        }
+        self.buf[self.pos..end].copy_from_slice(data);
+        self.pos = end;
+        Ok(())
+    }
+}
```
chrysh, line 61: `minicbor` already ships this writer: `encode::write::Cursor<&mut [u8]>` tracks
the position and returns `EndOfSlice` on overflow. Replacing `BufWriter` with an
alias keeps every `BufWriter::new(...)` call site working.

Three follow-ups after applying: change the four `w.pos` (lines 106, 189, 206,
223) to `w.position()`, drop `use minicbor::encode::Write as CborWrite;` on line
22, which is then unused, and apply the line-95
suggestion, which assumes this one has landed.

```suggestion
/// Writer over a fixed `[u8]` slice; tracks how many bytes have been written.
type BufWriter<'a> = minicbor::encode::write::Cursor<&'a mut [u8]>;
```

services/attest/producer/src/builder.rs
```diff
@@ -0,0 +1,428 @@
...
+        e.map(2).map_err(|_| AttestError::Cbor)?;
+        e.i64(1).map_err(|_| AttestError::Cbor)?;
+        e.i64(ALG_ES384).map_err(|_| AttestError::Cbor)?;
+        e.i64(HDR_X5CHAIN).map_err(|_| AttestError::Cbor)?;
+        e.array(chain.len() as u64).map_err(|_| AttestError::Cbor)?;
+        for cert in &chain {
+            e.bytes(cert).map_err(|_| AttestError::Cbor)?;
+        }
+        w.pos
+    };
```
chrysh: Apply the `Cursor` suggestion on line 51 first; this block depends on it.

`.map_err(|_| AttestError::Cbor)?` appears about 45 times across the four encode
blocks. One closure per block collapses it to four. It also fixes the error
vocabulary: the writer returns an overflow error, and today that gets relabelled
as `Cbor`, so a caller who overruns a scratch buffer is told the CBOR is
malformed. `minicbor::encode::Error::is_write()` separates the two.

Needs `use minicbor::encode::write::EndOfSlice;` at the top. The same shape
applies to the payload, Sig_Structure and COSE_Sign1 blocks; if all four get it,
lift the closure body's error mapping into a `fn cbor_err(...) -> AttestError`.

```suggestion
    let phdr_len = (|| -> Result<usize, minicbor::encode::Error<EndOfSlice>> {
        let mut w = BufWriter::new(&mut phdr_scratch[..]);
        let mut e = Encoder::new(&mut w);
        e.map(2)?;
        e.i64(1)?;
        e.i64(ALG_ES384)?;
        e.i64(HDR_X5CHAIN)?;
        e.array(chain.len() as u64)?;
        for cert in &chain {
            e.bytes(cert)?;
        }
        Ok(w.position())
    })()
    .map_err(|e| if e.is_write() { AttestError::BufferFull } else { AttestError::Cbor })?;
```

services/attest/producer/src/builder.rs
```diff
@@ -0,0 +1,428 @@
...
+    let phdr_bytes = &phdr_scratch[..phdr_len];
+
+    // ── Encode CWT claims map ──────────────────────────────────────────────
+    let mut payload_scratch = [0u8; SCRATCH];
+    let payload_len = {
+        let mut w = BufWriter::new(&mut payload_scratch);
+        let n_claims = 11 + usize::from(!evidence_cbor.is_empty());
```
chrysh: The 11 is a hand-maintained count of claims emitted 60 lines below, and it has
already been wrong once (f39a0db0 "fix: correct n_claims count"). Naming it and
listing the claims puts the invariant where someone adding a claim will see it.
The robust version is to encode the claims into scratch first and count as you
go, but that is a bigger change.

```suggestion
        // Fixed claims: iss, iat, nonce, ueid, oemid, hwmodel, hwver, dbgstat,
        // sw-name, sw-version, measurements. Update when adding a claim below.
        const FIXED_CLAIMS: usize = 11;
        let n_claims = FIXED_CLAIMS + usize::from(!evidence_cbor.is_empty());
```

services/attest/producer/src/builder.rs
```diff
@@ -0,0 +1,428 @@
...
+        e.tag(minicbor::data::Tag::new(61))
+            .map_err(|_| AttestError::Cbor)?;
+        e.map(n_claims as u64).map_err(|_| AttestError::Cbor)?;
+
+        e.i64(CLAIM_ISS).map_err(|_| AttestError::Cbor)?;
+        e.str("https://openprot.example/caliptra/device")
+            .map_err(|_| AttestError::Cbor)?;
+
+        e.i64(CLAIM_IAT).map_err(|_| AttestError::Cbor)?;
+        e.i64(iat as i64).map_err(|_| AttestError::Cbor)?;
```
chrysh: `iat` is a `u64` and CBOR has an unsigned major type, so the cast is not needed.
Same wire bytes for any real timestamp, so this is readability only.

```suggestion
        e.u64(iat).map_err(|_| AttestError::Cbor)?;
```

chrysh (commented):
Follow-up on the earlier review: 13 inline suggestions on style and API shape, all click-to-apply.

Three of them depend on another: comment on `builder.rs:25` needs the `builder.rs:82` slice change, `builder.rs:95` needs the `Cursor` alias on `builder.rs:51`, and `signer.rs:24` needs `MAX_PROVIDERS` in `consts.rs`. Each says so in its own text.

## Commit history

20+ commits mixing `fix:`, `deps:`, `build:` and `style:` prefixes that do not
match this repo's plainer component-first style (`orchestrator: ...`, `Add X`).
Several are fixups against this PR's own earlier commits. Not blocking if this
is squash-merged, but the squash message should be a single descriptive one.

services/attest/producer/src/signer.rs
```diff
@@ -0,0 +1,149 @@
...
+        &mut self,
+        provider: &'a dyn MeasurementProvider,
+    ) -> Result<(), AttestError> {
+        self.providers
+            .push(provider)
+            .map_err(|_| AttestError::BufferFull)
+    }
+}
+
+impl<'a> AttestProducer for HwAttestProducer<'a> {
```
chrysh: Unused lifetime parameter; `'_` is enough. (Line 27 does need `'a`, that one
stays.)

```suggestion
impl AttestProducer for HwAttestProducer<'_> {
```

services/attest/api/src/consts.rs
```diff
@@ -0,0 +1,20 @@
+// Licensed under the Apache-2.0 license
+// SPDX-License-Identifier: Apache-2.0
+
+// Caliptra chain: VendorCA → IDevID → LDevID → AliasFMC → AliasRT (leaf)
+pub const MAX_CHAIN_LEN: usize = 5;
+pub const MAX_CERT_SIZE: usize = 2048;
+
+pub const MAX_MEASUREMENTS: usize = 16;
```
chrysh, line 8: Name the provider capacity here, next to the other bounds.

```suggestion
pub const MAX_MEASUREMENTS: usize = 16;
pub const MAX_PROVIDERS: usize = 8;
```

services/attest/producer/src/signer.rs
```diff
@@ -0,0 +1,149 @@
...
+
+use crate::{builder, dice_identity, measurements};
+
+// ── Hardware-backed producer ──────────────────────────────────────────────────
+
+/// Attestation producer backed by a platform hardware signer.
+pub struct HwAttestProducer<'a> {
+    signer: &'a dyn HwSigner,
+    config: AttestConfig,
+    providers: Vec<&'a dyn MeasurementProvider, 8>,
```
chrysh: Bare 8 where the rest of the crate uses named `MAX_*` constants. Pairs with the
`MAX_PROVIDERS` suggestion in `consts.rs`; also add `MAX_PROVIDERS` to the
consts import on line 11.

```suggestion
    providers: Vec<&'a dyn MeasurementProvider, MAX_PROVIDERS>,
```

services/attest/producer/src/signer.rs
```diff
@@ -0,0 +1,149 @@
...
+            .map_err(|_| AttestError::BufferFull)?;
+        buf.push(leaf).map_err(|_| AttestError::BufferFull)?;
+        buf.push(ca).map_err(|_| AttestError::BufferFull)
+    }
+}
+
+// ── Stub signer used internally by SoftwareAttestProducer ────────────────────
+
+#[cfg(feature = "test-support")]
+struct StubSigner;
```
chrysh, line 149: The stub DER bytes `[0x30, 0x00]` are written out at twelve places across
`signer.rs`, `builder.rs` and `dice_identity.rs`. One constant here, used from
all of them, makes it obvious they are the same placeholder and gives it a name
to grep for when a real cert appears.

```suggestion
/// Placeholder DER: an empty SEQUENCE. Not a parseable certificate.
const STUB_CERT: [u8; 2] = [0x30, 0x00];

struct StubSigner;
```

third_party/crates_io/Cargo.toml
```diff
@@ -82,6 +82,8 @@ aes-gcm = { version = "0.10", default-features = false, features = ["aes"] }
 cipher = { version = "0.4", default-features = false }
 
 # Host tool crates needed by pigweed pw_kernel/tooling
+ciborium = { version = "0.2", default-features = false }
+minicbor = { version = "0.21", default-features = false }
```
chrysh, line 88: `minicbor` is a production `no_std` dependency of the attest producer, not host
tooling, so it should not sit under that comment. `ciborium` is test-only and can
stay where it is.

```suggestion
# Host tool crates needed by pigweed pw_kernel/tooling
ciborium = { version = "0.2", default-features = false }

# CBOR encoder for the attestation producer (no_std, production dependency)
minicbor = { version = "0.21", default-features = false }
```

gmandyam:
>producer/src/builder.rs build(): the ueid claim (key 256) is hard-coded to a constant placeholder (ueid[0] = 0x01, the remaining 28 bytes always zero) for every device, in the shared build() path used by both HwAttestProducer (production) and SoftwareAttestProducer. The README documents this claim as "Unique Entity Identifier from Caliptra CDI", but nothing derives it from device identity — every token from every device will carry the same UEID. This defeats the purpose of the claim and should either be wired to a real per-device source (e.g. via HwSigner/DICE identity) or clearly gated so it can't ship silently wrong (a // TODO alone is not enough given this is the production code path, not just the test stub).

ueid as I defined it in RFC 9711 (https://www.rfc-editor.org/info/rfc9711/#section-4.2.1) was meant to be determined directly from HW and not potentially overridden by a Caliptra cert ueid field (e.g. due to faulty implementation).  

On target, I expect the ueid to be available in HW (e.g. via fuse sense register), but we don't have a specific HW target.  That being said, I will modify the code to derive ueid from the cert chain, but I will also need to add a check to ensure that ueid is uniform across the entire cert chain.   I can also add a note in the readme that the implementor will have to replace this part of the code if ueid is determined from a source other than the Caliptra cert chain.

services/attest/producer/src/signer.rs
chrysh: Thanks, that covers signer.rs. The seven copies in the test modules are still
inline: builder.rs:236, 244, 246 and dice_identity.rs:46, 54, 70, 78. (The
0x30, 0x01 at dice_identity.rs:80 is a deliberately different cert, leave it.)

STUB_CERT is currently `#[cfg(feature = "test-support")]` and private, so those
modules cannot reach it. To share it, widen the gate and the visibility:

    #[cfg(any(test, feature = "test-support"))]
    pub(crate) const STUB_CERT: [u8; 2] = [0x30, 0x00];

then `use crate::signer::STUB_CERT;` in both test modules.

The [0x30, 0x00] in cert_ueid.rs are DER field placeholders, not stub certs, so
they stay as they are.

gmandyam: Please review.

STUB_CERT is now pub(crate) under #[cfg(any(test, feature = "test-support"))] in signer.rs, and both builder.rs and dice_identity.rs test modules import and use it — seven inline [0x30, 0x00] literals replaced.

services/attest/api/src/types.rs
```diff
@@ -0,0 +1,78 @@
...
+    fn sign(&self, payload: &[u8]) -> Result<[u8; 96], AttestError>;
+    /// Return the DER-encoded leaf certificate.
+    fn leaf_cert_der(&self, buf: &mut Vec<u8, MAX_CERT_SIZE>) -> Result<(), AttestError>;
+    /// Return the full DER-encoded certificate chain, leaf → root.
+    fn cert_chain_der(
+        &self,
+        buf: &mut Vec<Vec<u8, MAX_CERT_SIZE>, MAX_CHAIN_LEN>,
+    ) -> Result<(), AttestError>;
+    /// Return Caliptra-internal firmware measurements (ROM, FMC, runtime, etc.).
+    fn caliptra_measurements(
```
chrysh, line 61: Is this trait always going to be calling into caliptra? Otherwise, should we choose a more generic name? E.g. `rot_measurements`?

services/attest/api/src/types.rs
```diff
@@ -0,0 +1,78 @@
...
+        &self,
+        out: &mut Vec<Measurement, MAX_MEASUREMENTS>,
+    ) -> Result<(), AttestError>;
+}
+
+/// Platform-specific measurement source.
+///
+/// Implement for each firmware component (UEFI, BMC, etc.) the platform
+/// wants to measure beyond Caliptra-internal measurements.
+pub trait MeasurementProvider: Send + Sync {
```
chrysh: Why do we need Send + Sync here? Also check in all other places.

services/attest/producer/src/measurements.rs
```diff
@@ -0,0 +1,172 @@
...
+//! platform-registered [`MeasurementProvider`] instances.
+
+use heapless::Vec;
+
+use openprot_attest_api::consts::MAX_MEASUREMENTS;
+use openprot_attest_api::{AttestError, Measurement, MeasurementProvider};
+
+/// Collect all measurements: `caliptra` first, then registered providers.
+pub fn collect(
+    caliptra: &[Measurement],
```
chrysh: Are measurements caliptra specific? If not, can this be named more generic?

services/attest/api/src/consts.rs
```diff
@@ -0,0 +1,21 @@
...
+pub const MAX_PROVIDERS: usize = 8;
+pub const MAX_COMPONENT_LEN: usize = 64;
+pub const MAX_VERSION_LEN: usize = 32;
+pub const MAX_DIGEST_LEN: usize = 64; // SHA-512
+
+pub const MAX_OEMID_LEN: usize = 16;
+pub const MAX_HW_MODEL_LEN: usize = 64;
+pub const MAX_HW_VERSION_LEN: usize = 32;
+pub const MAX_NONCE_LEN: usize = 64;
```
chrysh: Unused param?

services/attest/api/src/consts.rs
```diff
@@ -0,0 +1,21 @@
...
+pub const MAX_EVIDENCE_LEN: usize = 4096;
```
chrysh: Unused param?

services/attest/producer/src/builder.rs
```diff
@@ -0,0 +1,423 @@
...
+
+    use crate::signer::STUB_CERT;
+
+    struct TestSigner;
+
+    impl HwSigner for TestSigner {
+        fn sign(&self, _: &[u8]) -> Result<[u8; 96], AttestError> {
+            Ok([0u8; 96])
+        }
+        fn leaf_cert_der(&self, buf: &mut Vec<u8, MAX_CERT_SIZE>) -> Result<(), AttestError> {
```
chrysh: leaf_cert_der has no callers. The leaf is already chain[0] after cert_chain_der, so every HwSigner implements a second mailbox path that nothing exercises. Drop it, or give it a  default body that calls cert_chain_der and takes the first entry, or name the caller you  have in mind.

services/attest/producer/src/measurements.rs
```diff
@@ -0,0 +1,172 @@
...
+    };
+    use openprot_attest_api::{DigestAlgorithm, MeasurementAuthority};
+
+    struct StubProvider {
+        name: &'static str,
+        fail: bool,
+    }
+
+    impl MeasurementProvider for StubProvider {
+        fn component_name(&self) -> &str {
```
chrysh: component_name has no callers, and Measurement already carries component, so the name is in the data every provider returns from measurements(). It is also the wrong shape: a provider that reports several components has one name for all of them. Drop it, or name the caller you have in mind.

services/attest/producer/src/measurements.rs
```diff
@@ -0,0 +1,172 @@
...
+use openprot_attest_api::{AttestError, Measurement, MeasurementProvider};
+
+/// Collect all measurements: `caliptra` first, then registered providers.
+pub fn collect(
+    caliptra: &[Measurement],
+    providers: &[&dyn MeasurementProvider],
+    out: &mut Vec<Measurement, MAX_MEASUREMENTS>,
+) -> Result<(), AttestError> {
+    for m in caliptra {
+        out.push(m.clone()).map_err(|_| AttestError::BufferFull)?;
```
chrysh: why do we use different conventions here? out.push vs p.measurements(out)?

Furthermore, "append" is a convention, not a contract. Nothing in the
  trait doc at types.rs:77-81 says a provider must only append. A provider calling
  out.clear() silently drops the RoT's own measurements, and the token still signs and verifies fine. That is a platform component erasing the root of trust's measurements from an attestation token.

Possible but maybe not the best fix in collect, but no extra buffer and no extra stack:

  for p in providers {
      let start = out.len();
      p.measurements(out)?;
      if out.len() < start {
          return Err(AttestError::Provider("provider shrank the measurement list"));
      }
  }

  plus one line on the trait: implementations append, they never remove or reorder existing entries.

services/attest/producer/src/measurements.rs
```diff
@@ -0,0 +1,172 @@
...
+                &p2 as &dyn MeasurementProvider,
+            ],
+            &mut out,
+        )
+        .unwrap();
+        assert_eq!(out.len(), 2);
+    }
+
+    #[test]
+    fn failing_provider_propagates_error() {
```
chrysh, line 157: This test only covers a provider that fails before pushing anything. A provider that pushes one entry and then fails leaves out half-populated. generate_token propagates the error so no token is emitted, so nothing leaks, but the buffer state on error is undefined and untested.

chrysh:
@gmandyam Are the tests already running in the CI when you push to the PR?

chrysh (changes requested):

gmandyam:
> @gmandyam Are the tests already running in the CI when you push to the PR?

CI tests are always re-executed upon a new push (as far as I can tell).  Is this not your experience?

services/attest/api/src/types.rs
gmandyam: This trait leverages Caliptra.  Making the name space generic is not going to change the dependencies on Caliptra.

services/attest/producer/src/measurements.rs
gmandyam: > Are measurements caliptra specific? If not, can this be named more generic?

These measurements are Caliptra-specific and are reflected in the token.

services/attest/producer/src/measurements.rs
gmandyam: > component_name has no callers, and Measurement already carries component, so the name is in the data every provider returns from measurements(). It is also the wrong shape: a provider that reports several components has one name for all of them. Drop it, or name the caller you have in mind.

Removed on latest push.

services/attest/api/src/consts.rs
gmandyam: > Unused param?

These are remnant const's that are no longer needed after moving to heapless.  Removed in latest push.

rusty1968:
@gmandyam remove the SKILLS from this PR. I created them for our internal use, it seems you added to your PR by accident.

services/attest/api/src/consts.rs
gmandyam: MAX_NONCE_LEN and MAX_EVIDENCE_LEN have been removed.  MAX_HW_MODEL_LEN and MAX_HW_VERSION_LEN are attestable constants, but HW_VERSION has now been removed as it is not required by OCP-EAT.

rusty1968:
Is the name of this service attest or verifier?

services/attest/producer/src/builder.rs
gmandyam: > leaf_cert_der has no callers. The leaf is already chain[0] after cert_chain_der, so every HwSigner implements a second mailbox path that nothing exercises. Drop it, or give it a default body that calls cert_chain_der and takes the first entry, or name the caller you have in mind.

Removed.

services/attest/api/README.md
```diff
@@ -0,0 +1,116 @@
...
+trait via the Caliptra mailbox driver (`caliptra-sw`).
+
+### `MeasurementProvider`
+
+Plug in platform-specific firmware measurement sources (UEFI, BMC, etc.)
+beyond the Caliptra-internal measurements.
+
+```rust
+pub trait MeasurementProvider: Send + Sync {
+    fn component_name(&self) -> &str;
```
chrysh: 784ab7a4 removed `Send + Sync`, `leaf_cert_der` and `component_name` from the code, but this README still documents all three: `HwSigner: Send + Sync` with `leaf_cert_der` at line 67, and `MeasurementProvider: Send + Sync` with `component_name` at line 89. Drop the bounds and the two methods so the README matches the traits.

services/attest/api/src/types.rs
chrysh: Can we guarantee that in 5 years caliptra will still be our only measurement source? If yes, then keep the name. If we might have a different measurement device, the name should be more generic.

Note the api crate itself already has no Caliptra dependency (BUILD deps are heapless and thiserror; the mailbox code is in the producer). So the name is the only Caliptra thing in the seam, and renaming it costs nothing.

gmandyam: > Can we guarantee that in 5 years caliptra will still be our only measurement source? If yes, then keep the name. If we might have a different measurement device, the name should be more generic.
> 
> Note the api crate itself already has no Caliptra dependency (BUILD deps are heapless and thiserror; the mailbox code is in the producer). So the name is the only Caliptra thing in the seam, and renaming it costs nothing.

OpenPRoT is intended to be an OCP standard (see https://www.amd.com/en/blogs/2025/openprot--building-a-secure-and-transparent-foundation-for-platf.html).  OCP created Caliptra and still refers normatively to Caliptra in relevant specifications (e.g. see https://opencomputeproject.github.io/Security/device-identity-provisioning/HEAD/).  The RoTs are not interchangeable from that perspective.  At very least any RoT must support TCG DICE and be OCP SAFE certifiable (https://github.com/opencomputeproject/OCP-Security-SAFE/tree/main/Reports/CHIPS_Alliance).

services/attest/api/README.md
gmandyam: > [784ab7a](https://github.com/OpenPRoT/openprot/commit/784ab7a4de981408e0d3f7512f5d00b0d015c039) removed `Send + Sync`, `leaf_cert_der` and `component_name` from the code, but this README still documents all three: `HwSigner: Send + Sync` with `leaf_cert_der` at line 67, and `MeasurementProvider: Send + Sync` with `component_name` at line 89. Drop the bounds and the two methods so the README matches the traits.

READMEs modified.  Will continue to check them before merge.

services/attest/producer/src/measurements.rs
gmandyam: > This test only covers a provider that fails before pushing anything. A provider that pushes one entry and then fails leaves out half-populated. generate_token propagates the error so no token is emitted, so nothing leaks, but the buffer state on error is undefined and untested.

Added new test rather than modifying existing test.

services/attest/producer/src/measurements.rs
gmandyam: >why do we use different conventions here? out.push vs p.measurements(out)?

Agentic inefficiency.  Change made.

>Furthermore, "append" is a convention, not a contract. Nothing in the
trait doc at types.rs:77-81 says a provider must only append. A provider calling
out.clear() silently drops the RoT's own measurements, and the token still signs and verifies fine. That is a platform component erasing the root of trust's measurements from an attestation token.

This is a reasonable suggestion, but I won't incorporate it for now.

OCP-EAT ((https://opencomputeproject.github.io/Security/ietf-eat-profile/HEAD/) is a profile of the EAT specification (RFC 9711).  I deliberately did not specify where profile compliance is enforced in RFC 9711 (https://www.rfc-editor.org/info/rfc9711/#section-6), so the enforcement point could be at the attester or verifier.

OCP-EAT requires certain mandatory claims including measurements.  However, the front matter of OCP-EAT states the following regarding mandatory claims:  "These claims are REQUIRED for all attestations and provide the minimum necessary information for verifier appraisal policies. The verifier can expect at a minimum these claims in a compliant attestation ...".  

It is up to the remote verifier to evaluate attestation tokens and make appraisal decisions based on the received evidence (including the profile claim), and this involves more than just verifying the token signature ("silently drops the RoT's own measurements, and the token still signs and verifies fine").  If the provider attempts to remove measurements that are expected in the token, the verification will fail because expected measurement values are missing.

gmandyam:
> @rusty1968 @gmandyam Are we adding AI skills to the repo now?

Removed the skills.

gmandyam:
> @gmandyam remove the SKILLS from this PR. I created them for our internal use, it seems you added to your PR by accident.

Done.

gmandyam:
> Is the name of this service attest or verifier?

Attest.  Verifier (local verifier) is part of the attestation service per design specification - see https://github.com/OpenPRoT/openprot/blob/main/docs/src/specification/services/attestation.md.

services/attest/producer/BUILD.bazel
```diff
@@ -0,0 +1,38 @@
...
+    name = "attest_producer_integration_test",
+    srcs = ["tests/producer_integration.rs"],
+    edition = "2021",
+    features = ["test-support"],
+    deps = [
+        ":attest_producer",
+        "//services/attest/api:attest_api",
+        "@rust_crates//:ciborium",
+    ],
+)
```
chrysh, line 38: `features` is Bazel's own attribute, not the rules_rust one. The Rust feature
never gets set, so `#![cfg(feature = "test-support")]` at
tests/producer_integration.rs:6 strips the whole file and the target compiles to
an empty crate. Bazel then reports it as PASSED with zero tests, locally and in
`./pw ci`, so all nine tests in there have never run.

Three changes to make them run. The attribute rename alone is not enough,
because `SoftwareAttestProducer` is itself behind the feature, so the test needs
a library built with it:

```suggestion
rust_library(
    name = "attest_producer_test_support",
    srcs = glob(["src/**/*.rs"]),
    crate_features = ["test-support"],
    crate_name = "openprot_attest_producer",
    edition = "2021",
    deps = [
        "//services/attest/api:attest_api",
        "@rust_crates//:heapless",
        "@rust_crates//:minicbor",
    ],
)

rust_test(
    name = "attest_producer_unit_test",
    crate = ":attest_producer",
    crate_features = ["test-support"],
    deps = [
        "//services/attest/api:attest_api",
    ],
)

rust_test(
    name = "attest_producer_integration_test",
    srcs = ["tests/producer_integration.rs"],
    crate_features = ["test-support"],
    edition = "2021",
    deps = [
        ":attest_producer_test_support",
        "//services/attest/api:attest_api",
        "@rust_crates//:ciborium",
        "@rust_crates//:heapless",
    ],
)
```

`heapless` is new in the test deps: it is used at producer_integration.rs:8 and
only shows up as an unresolved import once the file actually compiles.

With this on 73f549d6 the integration test runs 9 tests and all pass, the unit
test stays at 16. Checked with

    bazel test //services/attest/producer:attest_producer_integration_test \
        --test_output=all --nocache_test_results

chrysh: Filed the CI side as #462: a rust_test whose sources are all stripped by cfg reports PASSED with zero tests, so this target was green from the day it was added.

chrysh:
Looking at the CI log, those tests don't show up:

```
sign_receives_cose_sig_structure_for_payload
generate_token_rejects_chain_with_fewer_than_two_certs
nonce_outside_rfc9711_length_range_is_rejected
dbgstat_ueid_oemid_and_sw_claims_match_config
seventeenth_measurement_overflows_at_add_provider
provider_clearing_out_cannot_produce_a_token
```

The second one is not `dice_identity::tests::rejects_chain_with_fewer_than_two_certs`,
which does run: that covers the helper, and nothing covers the `generate_token`
path, which never calls it.

services/attest/api/src/types.rs
chrysh: The SAFE and DICE requirement is the part that argues for the generic name. "Any
RoT must support TCG DICE and be OCP SAFE certifiable" describes a class of
devices, so the trait method should be named after the role in that class, not
after one product in it. A normative reference to Caliptra says Caliptra is a
conforming RoT, not that a portable trait method has to carry its name.

This tree already ships more than one RoT. `target/earlgrey` is OpenTitan Earl
Grey, `target/veer` is the Caliptra emulator, and both run in `./pw ci`. On the
Earl Grey port the firmware measurements come from OpenTitan, not from a
Caliptra mailbox, so `HwSigner::caliptra_measurements` is already the wrong name
for one of the two platforms we build today.

To be clear about scope, I am not asking to scrub Caliptra from the crate.
`MeasurementAuthority::Caliptra` should stay: it tells the verifier the digest
came from Caliptra rather than from a platform provider asserting about itself,
which is exactly the distinction that decides how much the digest is worth.
`AttestError::Caliptra` for mailbox errors is fine too. Those name a real thing.
The trait method names a role, and that is the only line I want changed:

    fn caliptra_measurements(...)  ->  fn rot_measurements(...)

One rename plus its call sites, cheap now and awkward once anyone outside this
repo implements `HwSigner`.

One small ask on process: could you leave threads open until whoever raised
them has had a look? I resolve mine as soon as I have checked the fix, so
nothing sits around, and it keeps the open threads as an accurate list of what
is still being discussed.

services/attest/producer/BUILD.bazel
gmandyam: Thank you for pointing this out.  I will make the changes and verify the CI picks them up.

services/attest/api/src/types.rs
rusty1968: Agree with Chrsina on the naming - +1

> The SAFE and DICE requirement is the part that argues for the generic name. "Any RoT must support TCG DICE and be OCP SAFE certifiable" describes a class of devices, so the trait method should be named after the role in that class, not after one product in it. A normative reference to Caliptra says Caliptra is a conforming RoT, not that a portable trait method has to carry its name.
> 
> This tree already ships more than one RoT. `target/earlgrey` is OpenTitan Earl Grey, `target/veer` is the Caliptra emulator, and both run in `./pw ci`. On the Earl Grey port the firmware measurements come from OpenTitan, not from a Caliptra mailbox, so `HwSigner::caliptra_measurements` is already the wrong name for one of the two platforms we build today.
> 
> To be clear about scope, I am not asking to scrub Caliptra from the crate. `MeasurementAuthority::Caliptra` should stay: it tells the verifier the digest came from Caliptra rather than from a platform provider asserting about itself, which is exactly the distinction that decides how much the digest is worth. `AttestError::Caliptra` for mailbox errors is fine too. Those name a real thing. The trait method names a role, and that is the only line I want changed:
> 
> ```
> fn caliptra_measurements(...)  ->  fn rot_measurements(...)
> ```
> 
> One rename plus its call sites, cheap now and awkward once anyone outside this repo implements `HwSigner`.
> 
> One small ask on process: could you leave threads open until whoever raised them has had a look? I resolve mine as soon as I have checked the fix, so nothing sits around, and it keeps the open threads as an accurate list of what is still being discussed.

I agree the naming should reflect the vocabulary of the domain. There was another instance where Caliptra was part of the name of the type and we got rid of that.

gmandyam: You both are missing the main point:  the checks in the code (e.g. DICE identity) are specific to Caliptra.  That is by design. Changing the function naming is syntactic sugar - nothing more.   OpenTitan will fail as a DICE identity provider because it follows the OpenDICE standard.  If an IHV wants to use a different RoT then that is up to them to customize the code base. 

I will genericize the names but that won't affect the checks in the code.

chrysh: If it is caliptra specific, then it should not be in something called `services/attest/api/src/types.rs`. A programmer that looks at this directory expects something generic. Can you make sure that any code that is 
caliptra specific resides in a directory or filename that contains caliptra, e.g. services/attest/implementations/caliptra.rs? Any code that can also be reused for other attestation devices can be in the types.rs.

If we don't design our software this genric way, we will accumulate technical debt that we don't want for the future, which will make it harder to extend the software.

Does that make sense to you?

gmandyam: >Can you make sure that any code that is caliptra specific resides in a directory or filename that contains caliptra, e.g. services/attest/implementations/caliptra.rs?

Yes - this suggestion makes sense.

>If we don't design our software this genric way, we will accumulate technical debt that we don't want for the future, which will make it harder to extend the software.

Agree.

gmandyam: Change has been made by adding caliptra.rs. Please verify.

gmandyam:
> Looking at the CI log, those tests don't show up:
> 
> ```
> sign_receives_cose_sig_structure_for_payload
> generate_token_rejects_chain_with_fewer_than_two_certs
> nonce_outside_rfc9711_length_range_is_rejected
> dbgstat_ueid_oemid_and_sw_claims_match_config
> seventeenth_measurement_overflows_at_add_provider
> provider_clearing_out_cannot_produce_a_token
> ```
> 
> The second one is not `dice_identity::tests::rejects_chain_with_fewer_than_two_certs`, which does run: that covers the helper, and nothing covers the `generate_token` path, which never calls it.

Looks like the tests are showing up in the log now.

services/attest/api/src/signing_abstraction.rs
```diff
@@ -0,0 +1,114 @@
...
+        })
+    }
+}
+
+impl HwSigner for SwSigner {
+    /// Returns a zeroed 96-byte signature (r‖s placeholder).
+    ///
+    /// Production use must replace this with a real ECDSA P-384 implementation
+    /// using the stored private scalar.
+    fn sign(&self, _payload: &[u8]) -> Result<[u8; 96], AttestError> {
```
rusty1968: `SwSigner::sign()` always returns a hardcoded all-zero signature instead of a real ECDSA P-384 signature over the payload. `SwAttestProducer` (which uses `SwSigner`) is documented as a production path, not gated behind test-support. A platform integrator following the README's "SwAttestProducer (production, software key)" guidance would ship devices whose attestation tokens all carry a 96-byte all-zero signature — structurally valid but cryptographically meaningless — with nothing in the type system or build config preventing this from reaching production.

services/attest/producer/src/signer.rs
```diff
@@ -0,0 +1,278 @@
...
+        nonce: &[u8],
+        out: &mut Vec<u8, MAX_TOKEN_SIZE>,
+    ) -> Result<(), AttestError> {
+        let mut chain: Vec<Vec<u8, MAX_CERT_SIZE>, MAX_CHAIN_LEN> = Vec::new();
+        self.signer.cert_chain_der(&mut chain)?;
+
+        // Attempt UEID extraction from the leaf cert; fall back to a zeroed
+        // placeholder if the TCG UEID extension is absent.
+        let ueid: [u8; cert_ueid::UEID_LEN] =
+            cert_ueid::extract_and_verify(&chain).unwrap_or([0u8; cert_ueid::UEID_LEN]);
```
rusty1968: `SwAttestProducer::generate_token` swallows every error from `extract_and_verify` (including a genuine UEID mismatch across the certificate chain) via `unwrap_or`, silently substituting a zeroed UEID. If the leaf and a non-leaf cert in the SW-configured chain carry TCG UEID extensions with different values — exactly the tamper case `extract_and_verify` was written to catch — the resulting `AttestError::Mailbox("UEID mismatch across certificate chain")` is discarded and a token is generated anyway with a zero-filled UEID, instead of failing generation.

services/attest/producer/src/cert_ueid.rs
```diff
@@ -0,0 +1,391 @@
...
+    // cert DER: SEQUENCE { TBSCertificate, AlgorithmIdentifier, BIT STRING }
+    let tbs = sequence_body(cert_der).ok_or(AttestError::Mailbox("cert: bad outer SEQUENCE"))?;
+
+    // TBSCertificate: SEQUENCE { version [0], serialNumber, signature, issuer,
+    //                            validity, subject, spki, [3] extensions }
+    let tbs_body = sequence_body(tbs).ok_or(AttestError::Mailbox("cert: bad TBS SEQUENCE"))?;
+
+    // Walk TBS fields to find the [3] EXPLICIT extensions wrapper (tag 0xa3).
+    let extensions_wrapper =
+        find_tag(tbs_body, 0xa3).ok_or(AttestError::Mailbox("cert: no extensions"))?;
```
rusty1968: `extract()` returns `Err` (not `Ok(None)`) when a certificate's TBS has no `[3]` extensions wrapper at all, contradicting its own doc comment and the tolerant contract described for root CA certs. A vendor root CA cert that legitimately carries no X.509v3 extensions section causes `find_tag(tbs_body, 0xa3)` to return `None`, which `.ok_or(...)?` turns into `AttestError::Mailbox` — propagated by `extract_and_verify`'s `extract(cert)?` in the `skip(1)` loop, and further propagated unguarded by `HwAttestProducer::generate_token`, so a legitimate hardware attestation chain fails entirely instead of the root cert being "skipped without error" as documented.

services/attest/producer/src/signer.rs
```diff
@@ -0,0 +1,278 @@
...
+        self.signer.measurements(&mut meas)?;
+        measurements::collect(&self.providers, &mut meas)?;
+        builder::build(&self.config, self.signer, &ueid, &meas, nonce, out)
+    }
+
+    fn cert_chain(
+        &self,
+        buf: &mut Vec<Vec<u8, MAX_CERT_SIZE>, MAX_CHAIN_LEN>,
+    ) -> Result<(), AttestError> {
+        dice_identity::cert_chain(self.signer).map(|c| *buf = c)
```
rusty1968: `AttestProducer::cert_chain` has inconsistent semantics across implementations: `HwAttestProducer` overwrites the caller's buffer (`*buf = c`) while `SwAttestProducer` and `SoftwareAttestProducer` append to it via `buf.push(...)` without clearing first. Code that reuses/pre-populates the buffer across calls, or calls `cert_chain()` twice on the same buffer, accumulates duplicate certificates on the software path and can eventually overflow `MAX_CHAIN_LEN` returning `BufferFull` where the Hw path would have succeeded.

services/attest/producer/src/builder.rs
```diff
@@ -0,0 +1,506 @@
...
+        e.i64(1)?;
+        e.i64(ALG_ES384)?;
+        Ok(w.position())
+    })()
+    .map_err(cbor_err)?;
+    let phdr_bytes = &phdr_scratch[..phdr_len];
+
+    // ── Encode CWT claims map ──────────────────────────────────────────────
+    // Claims are written in CBOR deterministic order (RFC 8949 §4.2.1).
+    let mut payload_scratch = [0u8; SCRATCH];
```
rusty1968: `build()` allocates three separate 8192-byte stack buffers (`payload_scratch`, `sig_scratch`, `cose_scratch`) plus a 16-byte header buffer on every token-generation call in a `#![no_std]` embedded crate — roughly 24.5KB of stack simultaneously live across nested closures. On a constrained microcontroller stack (Caliptra-class firmware typically has only a few KB of stack per task), this risks stack overflow/corruption on real hardware even though actual encoded content is usually far smaller than `MAX_TOKEN_SIZE`.

services/attest/api/BUILD.bazel
```diff
@@ -0,0 +1,21 @@
...
+    edition = "2021",
+    visibility = ["//visibility:public"],
+    deps = [
+        "@rust_crates//:heapless",
+        "@rust_crates//:thiserror",
+    ],
+)
+
+rust_test(
+    name = "attest_api_test",
```
rusty1968: The `attest_api_test` Bazel target exists but the `api` crate (`traits.rs`, `types.rs`, `signing_abstraction.rs`, `error.rs`, `consts.rs`) contains no `#[cfg(test)]` modules, so the security-critical `SwSigner::new` validation (zero-scalar check, P-384 order check, DER SEQUENCE check) has zero test coverage despite a passing test target implying otherwise. A future regression in that validation would pass silently.

services/attest/api/src/types.rs
```diff
@@ -0,0 +1,82 @@
...
+///
+/// `private_key_scalar` is the 48-byte big-endian P-384 private scalar `d`.
+/// It must satisfy `1 ≤ d < n` (P-384 group order) — validated at construction
+/// time by [`crate::SwSigner::new`].
+///
+/// `cert_chain` holds the DER-encoded certificate chain, leaf → root.  Each
+/// certificate must start with `0x30` (DER SEQUENCE).  At least one certificate
+/// is required.
+pub struct SwSignerConfig {
+    pub private_key_scalar: [u8; 48],
```
rusty1968: `SwSignerConfig`/`SwSigner` hold the raw 48-byte P-384 private key scalar in a plain struct with no zeroization on drop, leaving key material in stack/heap memory longer than necessary and widening the window for recovery via a memory-disclosure bug or cold-boot style attack on the embedded target.

services/attest/producer/src/signer.rs
```diff
@@ -0,0 +1,278 @@
...
+            providers: Vec::new(),
+        }
+    }
+
+    pub fn add_provider(
+        &mut self,
+        provider: &'a dyn MeasurementProvider,
+    ) -> Result<(), AttestError> {
+        self.providers
+            .push(provider)
```
rusty1968: `HwAttestProducer::generate_token` re-implements the fetch+validate cert-chain logic that already exists as `dice_identity::cert_chain()`, duplicating those two lines inline instead of reusing it. A future change to the validation sequence applied at only one of the two call sites will silently diverge between `generate_token` and `cert_chain()`.

third_party/crates_io/Cargo.toml
```diff
@@ -82,6 +82,10 @@ aes-gcm = { version = "0.10", default-features = false, features = ["aes"] }
 cipher = { version = "0.4", default-features = false }
 
 # Host tool crates needed by pigweed pw_kernel/tooling
+ciborium = { version = "0.2", default-features = false }
```
rusty1968: `minicbor`, described in its own comment as a "no_std, production dependency" for the attestation producer, is inserted directly under the "# Host tool crates needed by pigweed pw_kernel/tooling" section comment alongside `ciborium` (an actual host-only test dependency), misclassifying it in the file's own organizational scheme. A future contributor could assume everything in that section is host-only and exclude `minicbor` from embedded production builds.

rusty1968 (changes requested):
Automated review findings (most severe first). See inline comments.

services/attest/api/src/types.rs
```diff
@@ -0,0 +1,82 @@
...
+    Hardware,
+    /// Software-only signing with caller-supplied key material.
+    Software,
+}
+
+/// Producer configuration, set once at platform initialisation.
+pub struct AttestConfig {
+    pub oemid: OemId,
+    pub hw_model: String<MAX_HW_MODEL_LEN>,
+    pub cert_cache_ttl: Duration,
```
rusty1968: `cert_cache_ttl` and `signer_kind` (next line) are never read anywhere in this PR — `builder::build()` only consumes `config.oemid` and `config.hw_model`. A caller populating `cert_cache_ttl` would reasonably assume it drives caching behavior; it currently does nothing. `signer_kind` also duplicates information already fully determined by which concrete producer type (`HwAttestProducer` vs `SwAttestProducer`) the caller constructs, and nothing prevents constructing e.g. an `HwAttestProducer` with `signer_kind: SignerKind::Software` — a config/reality mismatch the type system should make impossible rather than silently allow. Recommend either wiring these fields to real behavior before merge, or dropping them until they're needed.

services/attest/api/src/error.rs
```diff
@@ -0,0 +1,18 @@
+// Licensed under the Apache-2.0 license
+// SPDX-License-Identifier: Apache-2.0
+
+#[derive(Debug, thiserror::Error)]
+pub enum AttestError {
+    #[error("Mailbox error: {0}")]
+    Mailbox(&'static str),
```
rusty1968: `Mailbox(&'static str)` reads as "Caliptra mailbox communication failure" but is reused as a generic string-bucket for unrelated failure domains: DER parsing (`cert_ueid.rs:42`, e.g. `"cert: bad outer SEQUENCE"`), DICE chain validation (`dice_identity.rs`, e.g. `"cert: not X.509 v3"`), and even nonce-length validation (`builder.rs:80`, `"nonce must be 8–64 bytes..."`) — none of which touch a mailbox. Callers can't `match` on failure category, only inspect an opaque string. Worth splitting into a real taxonomy (e.g. `Der`, `InvalidNonce`, `ChainValidation`, `Mailbox`) before external callers start pattern-matching on this enum.

services/attest/producer/src/cert_ueid.rs
```diff
@@ -0,0 +1,391 @@
+// Licensed under the Apache-2.0 license
```
rusty1968: This module's doc says "Minimal DER walker for extracting the TCG UEID" — narrow and UEID-specific. But `sequence_body`, `parse_length`, `take_sequence`, `octet_string_body`, `find_tag`, `is_x509_v3` (line 117), and `has_extension_oid` (line 128) are general DER/X.509 primitives, and `dice_identity.rs:18` imports two of them (`has_extension_oid`, `is_x509_v3`) for an unrelated purpose (DICE chain compliance, not UEID extraction). A reader looking for "the DER parser" won't find it under a module named for UEID extraction. Consider splitting the generic DER primitives into their own module (e.g. `der.rs`) so both modules are honestly named and independently reusable.

services/attest/api/src/traits.rs
```diff
@@ -0,0 +1,25 @@
...
+    /// Generate a signed OCP-EAT COSE_Sign1 token bound to `nonce`.
+    ///
+    /// The encoded token is appended to `out`.
+    fn generate_token(
+        &self,
+        nonce: &[u8],
+        out: &mut Vec<u8, MAX_TOKEN_SIZE>,
+    ) -> Result<(), AttestError>;
+
+    /// Return the current DICE certificate chain, ordered leaf → root.
```
rusty1968: This doc says "Return the current DICE certificate chain" but doesn't specify whether `buf`'s existing contents are cleared or appended to. The two implementations in the producer crate disagree: `HwAttestProducer::cert_chain` (`signer.rs:69`) overwrites via `*buf = c`, while `SwAttestProducer::cert_chain` appends via `push` without clearing. That behavior belongs in the trait contract, not left as an implementation detail callers discover by reading source — especially since a caller reusing/pre-populating `buf` across calls gets silently different results depending on which producer they hold.

services/attest/producer/src/builder.rs
```diff
@@ -0,0 +1,506 @@
...
+    let payload_len = (|| -> Result<usize, minicbor::encode::Error<EndOfSlice>> {
+        // Fixed claims (7): nonce, ueid, oemid, hwmodel, dbgstat,
+        // eat_profile, measurements. Update on any change below.
+        const FIXED_CLAIMS: usize = 7;
+        let n_claims = FIXED_CLAIMS;
```
rusty1968: `FIXED_CLAIMS: usize = 7` is a hand-maintained count synchronized only by the comment above it ("Update on any change below") against the sequence of `e.i64()/e.bytes()` calls that follow. Adding an 8th claim without updating this constant produces a token whose declared CBOR map length doesn't match its actual entry count — likely undetected here until a downstream decoder chokes on it. Consider deriving the count structurally (e.g. building the claims from an array/slice and using its length) rather than relying on a comment to keep two independently-edited things in sync.

rusty1968 (commented):
Design review (Ousterhout-style: module depth, information hiding, complexity distribution) as a follow-up to the correctness review above.

services/attest/producer/src/builder.rs
rusty1968: Concrete mitigation, roughly in order of effort/impact:

1. **Drop `cose_scratch` entirely (saves 8KB, no API change).** `out` is already caller-allocated at `MAX_TOKEN_SIZE`. Resize it to capacity and encode the final COSE_Sign1 directly into its backing slice, then truncate to the actual length, instead of building it in a throwaway stack array and copying:
   ```rust
   out.resize_default(MAX_TOKEN_SIZE).map_err(|_| AttestError::BufferFull)?;
   let cose_len = (|| -> Result<usize, _> {
       let mut w = BufWriter::new(&mut out[..]);
       let mut e = Encoder::new(&mut w);
       // ...same encoding as today...
       Ok(w.position())
   })().map_err(cbor_err)?;
   out.truncate(cose_len);
   ```

2. **Right-size `payload_scratch`/`sig_scratch`.** Both are sized at `MAX_TOKEN_SIZE` (8192) but real content is bounded by known consts (nonce ≤64, UEID 17, oemid ≤16, hw_model ≤64, a handful of small int claims, plus `MAX_MEASUREMENTS` records of `MAX_COMPONENT_LEN + MAX_VERSION_LEN + MAX_DIGEST_LEN`). A computed const (e.g. ~256 bytes fixed overhead + `MAX_MEASUREMENTS * (component+version+digest+per-entry overhead)`) lands closer to ~2KB than 8KB. `sig_scratch` only needs `payload size + ~64` bytes of `Sig_Structure` overhead, so it can shrink in step with `payload_scratch`.

Together, #1 and #2 take this from ~24.5KB to roughly 4-5KB with no interface change.

3. **If more headroom is still needed:** move the scratch buffers off the stack entirely — into the producer struct (allocated once at construction) instead of re-declared on every `generate_token()` call — so the cost is a fixed static/struct-owned region rather than added to whatever stack frame happens to call into `build()`. Only worth doing if profiling still shows pressure after #1/#2.

rusty1968: Separately from fixing this instance: worth adding a guard rail so this class of issue doesn't need a human/reviewer to catch it next time.

- **`clippy::large_stack_arrays`** (stable, no nightly needed) flags stack arrays over a size threshold, configurable via `clippy.toml`'s `array-size-threshold`. Setting that to something embedded-appropriate (e.g. 1-2KB) and running clippy in CI/Bazel's `rust_clippy` would have failed the build on the three `[0u8; 8192]` buffers here automatically, rather than relying on review to spot it.
- **`cargo-call-stack`** is the heavier option if worst-case *call-chain* stack depth matters (not just individual arrays) — it computes per-function stack usage on ARM Cortex-M targets and can catch a chain of small buffers that adds up across nested calls, which a single-array lint won't. More CI plumbing, and only works on embedded targets (not host unit tests).

Given this crate is headed for a Caliptra-class MCU with only a few KB of task stack, I'd start with the clippy lint (cheap, catches the common case) and consider `cargo-call-stack` later if stack overflows show up in practice despite it.

rusty1968: Correction to my clippy suggestion above — I checked this repo's actual Bazel setup before assuming Cargo-style config would apply, and it doesn't cleanly:

- There's no clippy wired into the build at all here — no `clippy.toml`, no `rust_clippy`/`rust_clippy_aspect` target anywhere (the only "rust_clippy" hits in the repo are prose in docs, not actual BUILD rules). This crate has no Cargo.toml either (per the earlier review thread asking to drop it).
- `clippy::large_stack_arrays`'s size *threshold* is only configurable via `clippy.toml`, which clippy discovers by walking up from `CARGO_MANIFEST_DIR` — a directory that doesn't exist in a Bazel-only crate. A crate-level `#![deny(clippy::large_stack_arrays)]` in `lib.rs` would still work (it's a rustc/clippy attribute, build-system-agnostic), but without a discoverable `clippy.toml` it only fires at the **default 512,000-byte threshold** — which would *not* have caught the 8KB buffers here. So my suggestion as stated wouldn't actually have worked in this repo without first standing up clippy-under-Bazel plumbing (new `rust_clippy_aspect` target, some way to get a custom `clippy.toml` path to `clippy-driver`) — a bigger lift than "tune a threshold," and not something I want to hand over unverified flag names for.

Given that, a lower-effort guard rail that actually matches this repo's existing conventions: there's already a house-style presubmit check pattern here (`presubmit/license.py`, `presubmit/cpp_include_guard.py`, wired through `presubmit/presubmit.py`). A similar small script that flags stack-array declarations over a size threshold in Rust sources would catch this class of bug without needing a new Bazel/clippy integration. Retracting the clippy-as-stated suggestion in favor of that.

Issue #377, opened 2026-07-28
Formalize the trial-boot / commit / rollback capabilities as traits


rusty1968:
fwmanager's BootControl doc already sketches this flow (set_trial → release → await_boot → commit), 

```rust
pub enum BootProgress { Booted, Failed, Timeout }

pub trait BootMonitor {
    type Error: core::error::Error;
    fn await_boot(&mut self, window: core::time::Duration)
        -> Result<BootProgress, Self::Error>;
}

pub trait SlotControl {
    type Error: core::error::Error;
    fn set_trial(&mut self, slot: Slot) -> Result<(), Self::Error>;
    fn commit(&mut self, slot: Slot)    -> Result<(), Self::Error>;
    fn rollback(&mut self)              -> Result<(), Self::Error>;
}


fn trial_boot<D, M, S>(dev: &mut D, monitor: &mut M, store: &mut S, new_slot: Slot)
    -> Result<(), Error>
where
    D: BootControl,
    M: BootMonitor,
    S: SlotStore,
{
    dev.hold_in_reset()?;
    store.set_trial(new_slot)?;                 // tentative — not yet committed
    dev.release()?;                             // boot the trial image

    match monitor.await_boot(window)? {         // Ok(progress); Err only = couldn't watch
        BootProgress::Booted  => store.commit(new_slot)?,   // observed good => active
        BootProgress::Failed  => store.rollback()?,         // bad image => previous slot
        BootProgress::Timeout => store.rollback()?,         // went quiet => previous slot
    }
    Ok(())
}
```

PR #392, opened 2026-08-04
fwmanager: downstream update lifecycle tests with slot and boot-progress capabilities

chrysh:
Host-level, black-box integration tests for the downstream firmware-update lifecycle: a MockDownstreamDevice on the eRoT's seams (reset line, boot-complete line, slot flash, staging flash, slot store), with every assertion made on externally visible signals — reset transitions, what was flashed while the device was held, which slot the device booted, what got committed. Plain rust_test, no kernel, no QEMU.

Covered scenarios: a successful update proves the device is held in reset the whole time firmware is written, is rebooted on the new slot, and is committed only after the booted state was observed; a device whose firmware write path fails is noticed through the flash seam, with nothing armed or committed and the old slot untouched; a corrupt staged image is rejected during staging without disturbing the running device; a trial-boot window that expires without evidence rolls back and reboots the still-committed slot. A negative-control test runs a deliberately careless driver and asserts the instrumentation catches it, so the positive assertions cannot pass vacuously.

Supporting pieces: the SlotControl capability (trial/commit/rollback slot selection) and BootProgress/await_boot (a bounded boot window over the existing BootMonitor, rendered as a deterministic poll budget — tests never sleep) are added to the fwmanager api leaf, per #377. The mock downstream device grows A/B slots, a staging region, a slot store, fault injection for a broken write path, and the ordering instrumentation the tests assert on. The existing boot_flow tests pass unchanged on the extended mock.

There is deliberately no orchestrator dependency: stage_update/activate_update in the test file are the reference rendering of the trial-boot protocol sketched in the BootControl docs. They are the executable spec for the orchestrator from #357 — when it merges, the real state machine replaces these test-local drivers and the assertions stay.

Refs #377.

### DIFF
--- a/platform/impls/baremetal/mock/BUILD.bazel
+++ b/platform/impls/baremetal/mock/BUILD.bazel
@@ -16,6 +16,7 @@ rust_library(
     deps = [
         "//hal/blocking",
         "//hal/blocking/flash",
+        "//services/fwmanager/api:fwmanager_api",
         "//util/types",
         "@rust_crates//:cortex-m",
         "@rust_crates//:embedded-hal",
--- a/platform/impls/baremetal/mock/BUILD.bazel
+++ b/platform/impls/baremetal/mock/BUILD.bazel
@@ -15,6 +15,8 @@ rust_library(
     edition = "2024",
     deps = [
         "//hal/blocking",
+        "//hal/blocking/flash",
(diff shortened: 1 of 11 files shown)


chrysh:
The test functions called are not up to date anymore since the orchestrator commit was merged.

PR #394, opened 2026-08-05
orchestrator: Add the per-device image layout to the device-table schema

chrysh:
Redesigned since the last review round: SlotDesc, RecoveryPolicy and the
per-slot flags are gone, so the earlier comments no longer apply.

Board tables now say where each device's images live. `ImageLayout` holds the
slots the eRoT writes plus the golden image; a `Slot` is an id and a `Region`, a
byte range counted from the start of that device's firmware partition.
`ImageLayout::new` rejects duplicate slot ids and overlapping regions at build
time.

Golden is its own type, not a flag on a slot, so code that picks a write target
or checks the SVN floor never sees it. The floor has to skip it: a golden
image's security version is fixed when the board is made, so enforcing the floor
would eventually make the last image that still boots unbootable.

The layout is optional. A device that takes its own updates over PLDM and picks
what it boots declares none.

`assert_retry_reaches_every_image` closes the gap rusty1968 raised: boards call
it from a const fence, so a retry budget too small to boot every image fails the
build. The floor is one more than the image count, because the last restore the
budget allows is never booted.

Part of 9elements/openprot#1.

### DIFF
--- a/services/orchestrator/config/src/lib.rs
+++ b/services/orchestrator/config/src/lib.rs
@@ -81,9 +81,12 @@ impl<G> BootCheckpoint<G> {
     }
 }
 
-/// Identifies one slot within one device's layout. Opaque: meaning comes
-/// from the position in that device's slot table, never from a global
-/// vocabulary — slot 0 on the BMC and slot 0 on the NIC are unrelated.
+/// Identifies one slot within one device's layout. An opaque per-device
+/// token, not an index: ids need only be unique within one device's table
+/// ([`DeviceConfig::new`] enforces exactly that) — they are not required
+/// to be contiguous, ordered, or to start at zero, and slot 0 on the BMC
+/// and slot 0 on the NIC are unrelated. Ladder order comes from table
+/// declaration order, never from id values.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub struct SlotId(pub u8);
 
--- a/services/orchestrator/config/src/lib.rs
+++ b/services/orchestrator/config/src/lib.rs
@@ -81,6 +81,41 @@ impl<G> BootCheckpoint<G> {
     }
 }
 
+/// Identifies one slot within one device's layout. Opaque: meaning comes
+/// from the position in that device's slot table, never from a global
+/// vocabulary — slot 0 on the BMC and slot 0 on the NIC are unrelated.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct SlotId(pub u8);
+
+/// A distinguished duty a slot carries beyond being writable/bootable.
+///
+/// Intentionally exhaustive (not `#[non_exhaustive]`): adding a role is a
+/// breaking change, so every consumer that dispatches on roles — in
+/// particular recovery-candidate selection — is forced to handle the new
+/// role explicitly instead of falling into a wildcard arm.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum SlotRole {
+    /// The slot recovery falls back to after every ordinary rung failed.
+    /// A "golden" image is this role plus `writable: false` — a property
+    /// combination, not a separate name.
+    Recovery,
+}
+
+/// One slot in a device's layout: topology as data, not a type. A layout
+/// is plain A/B, A/B + golden, or single + golden purely by what the table
+/// declares — no layout shape is named anywhere.
+#[derive(Debug, Clone, Copy)]
+pub struct Slot {
+    pub id: SlotId,
+    /// May the update path write this slot? `false` on a recovery-role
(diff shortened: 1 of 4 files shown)


services/orchestrator/config/src/lib.rs
```diff
@@ -97,6 +165,43 @@ pub const fn validate<R, G>(devices: &[DeviceConfig<R, G>]) {
...
+        let mut bootable = 0;
+        let mut recovery_slots = 0;
+        let mut s = 0;
+        while s < slots.len() {
+            if slots[s].bootable {
+                bootable += 1;
+            }
+            if matches!(slots[s].role, Some(SlotRole::Recovery)) {
+                recovery_slots += 1;
+                assert!(slots[s].bootable, "a recovery-role slot must be bootable");
```
leongross: To prevent invalid slot definitions with role `Recovery` and bootable `false` we could create a simple constructor that prohibits this and throws an error.  


```rust
type SlotDescError ...

impl SlotDesc {
	pub fn new(id: u8, writebale: bool, bootable: bool, policy: Option<SlotRole>) -> Result<SlotDescError, Self> {
    ...
  }
}
```

leongross (changes requested):

services/fwmanager/api/src/config.rs
```diff
@@ -53,6 +58,62 @@ pub struct BootCheckpoint<G> {
...
+    /// A "golden" image is this role plus `writable: false` — a property
+    /// combination, not a separate name.
+    Recovery,
+}
+
+/// One slot in a device's layout: topology as data, not a type. A layout
+/// is plain A/B, A/B + golden, or single + golden purely by what the table
+/// declares — no layout shape is named anywhere.
+#[derive(Debug, Clone, Copy)]
+pub struct SlotDesc {
```
leongross: Generally, we should discuss whether a SlotDesc should be mutable or not. If not, we can make validation a method for the SlotDesc and run it on creation to make the API safer for the users, such that the user cannot 'forget' to validate the SlotDesc.

chrysh: A constructor is a good idea, done

services/orchestrator/config/src/lib.rs
chrysh: a constructor is a good idea, done

services/orchestrator/config/src/lib.rs
```diff
@@ -75,6 +163,13 @@ pub struct DeviceConfig<R, G: 'static> {
     /// checkpoint whose window expires fails the boot.
     pub checkpoints: &'static [BootCheckpoint<G>],
     pub commit_policy: CommitPolicy,
+    /// This device's slot layout. The recovery ladder is derived from it,
+    /// never declared: bootable slots in declaration order, recovery-role
+    /// slot last, escalation to out-of-band recovery once no rung is left.
+    /// A layout without rungs — e.g. empty, for a device that owns its
+    /// boot selection internally (the PLDM archetype) — leaves escalation
+    /// as the only step.
+    pub slots: &'static [Slot],
```
rusty1968, line 122: A board can declare more recovery slots than the orchestrator is allowed to try!

How far recovery falls back is set by max_retry (in the orchestrator), but the ladder depth is set here (in the slot table), and nothing checks that they agree. 

If max_retry is smaller than the number of slots, the sm locks up before it ever reaches the golden slot — so the last-resort image silently never runs.

Since this is a schema-only PR, either add a build-time check that max_retry covers the deepest ladder or note it as a required follow-up before anything starts using these slots. 

Otherwise a board can ship with an unreachable golden slot and no one finds out.

rusty1968 (changes requested):

services/orchestrator/config/src/lib.rs
leongross: Where is this implemented @chrysh, I can't seem to find it in this patch.

services/orchestrator/config/src/lib.rs
chrysh: Why do we have max_retry ? Shouldn't it be dictated by the number of configured slots? The sm just tries the next configured slot until no more slots left to try?

chrysh:
Closed until @FerralCoder returns the config file and gives us more instructions.

services/orchestrator/config/src/lib.rs
chrysh: I changed it, now golden has its own role.

chrysh:
This commit is already on ocp-global-demo-wip as 739b2add, picked up when that branch was rebased.

It is not a clean copy of the one here. The golden image became a required field there, and the paragraph saying it is optional dropped out of the commit message. #493 puts `Option<Golden>` back on the demo branch, along with the check that rejects a layout holding no image at all.

PR #395, opened 2026-08-05
fwmanager: Add the boot-progress walk over checkpoint windows

chrysh:
**Stacked on #394** — the first three commits are that PR; review starts at `fwmanager: Add the orchestrator root crate with the boot-walk core`.

`BootWalk` is the first consumer of the fwmanager capability seams: after a device is released from reset, it walks the device's configured `BootCheckpoint`s and judges the boot good or failed. Windows are sequential (each opens when the previous checkpoint is observed), `Booted` at the deadline still advances, and a passive device (`CommitPolicy::None`) is complete from construction. The walk is sans-IO — time is injected as monotonic milliseconds — so every verdict is host-tested; the kernel-side pump comes with the orchestrator runtime.

Signal-to-monitor dispatch stays out of the api crate: a `MonitorMap` owned by the board wiring resolves each checkpoint's signal, so `BootMonitor` is untouched and one transport can back several signal kinds through per-signal views. `poll()` tolerates two consecutive monitor read errors (transports can glitch during bring-up); the third fails the walk as `MonitorRead` with the concrete error on the `source()` chain. An unwired signal fails as `UnmappedSignal` instead of hanging.

The end-to-end test runs the composition loop over the mock device table: declaration order is boot order, one device at a time, the passive cpld consumes no polls, and a missing heartbeat stops the walk naming `nic`/`heartbeat`. The `comes_up_within` placeholder in the api tests is retired in favor of the walker's tests.

Part of 9elements/openprot#3 (recovery initiation trigger) and 9elements/openprot#15 (boot-failure judgment for partition fallback).

### DIFF
--- a/services/fwmanager/BUILD.bazel
+++ b/services/fwmanager/BUILD.bazel
@@ -1,0 +1,23 @@
+# Licensed under the Apache-2.0 license
+# SPDX-License-Identifier: Apache-2.0
+
+load("@rules_rust//rust:defs.bzl", "rust_library", "rust_test")
+
+rust_library(
+    name = "fwmanager",
+    srcs = [
+        "src/boot_progress.rs",
+        "src/lib.rs",
+    ],
+    edition = "2024",
+    visibility = ["//visibility:public"],
+    deps = [
+        "//services/fwmanager/api:fwmanager_api",
+    ],
+)
+
+# Host tests: build on the host platform, no kernel/QEMU.
+rust_test(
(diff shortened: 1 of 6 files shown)


chrysh:
Dropping this: superseded by a redesign that removes BootMonitor entirely — checkpoints become board-table data (signal ids resolved by a per-device EvidenceReader), and devices can report failure and its retriability as evidence. A fresh traits-only PR follows from branch add-boot-walk.

PR #398, opened 2026-08-06
docs: refresh architecture map

Daniiiil1:
## What changed

- add the target-agnostic USART driver layer to the repository map
- document the current firmware services, including I2C, MCTP, SPDM, firmware management, and the orchestrator
- add AST10x0 and mock targets alongside Earl Grey and VeeR
- explain how interfaces, platform implementations, target backends, firmware images, and host/target tests are separated

## Why

The architecture guide no longer reflected several directories and abstractions that are present on `main`. Updating the map makes it easier for new contributors to find the correct layer for portable protocol logic, silicon-specific implementations, and system-image tests.

## Validation

- `git diff --check`
- `npx --yes @bazel/bazelisk build //docs` was attempted twice. The first run timed out fetching Pigweed LLVM through CIPD. The second completed analysis and 705/706 actions, then failed while linking the external mdBook tool because Pigweed’s macOS sysroot could not resolve the CFNetwork framework re-export from CoreServices. The failure occurs in the external documentation toolchain before processing this Markdown change.

### DIFF
--- a/docs/src/architecture.md
+++ b/docs/src/architecture.md
@@ -12,6 +12,8 @@ openprot/
 ├── workflows.json            # Pigweed workflow groups (./pw …)
 ├── BUILD.bazel               # Top-level Bazel package
 ├── openprot/                 # Main application crate (lib.rs + main.rs)
+├── drivers/
+│   └── usart/                # Target-agnostic USART API, client, and server
 ├── hal/
 │   ├── async/                # Async HAL trait crates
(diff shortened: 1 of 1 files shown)


rusty1968:
we have renamed fwmanager to orchestrator.

rusty1968 (changes requested):
the refresh needs a refresher.

rusty1968 (approved):

PR #401, opened 2026-08-06
docs: drop removed multilingual field from book.toml

chrysh:
mdbook 0.5 rejects the field; serve refused to start.

Assisted-by: Claude:claude-fable-5


leongross (approved):
lgtm

PR #403, opened 2026-08-06
services/orchestrator: split api crate into capabilities and config

rusty1968:
Rename orchestrator/api to orchestrator/capabilities (BootControl, BootMonitor ports) and move the device-table schema into a new orchestrator/config crate. Update hal-adapters, sm doc references, and the mock target's device table.


chrysh (approved):
Good idea!

Issue #26, opened 2025-09-04
Enable security vulnerability reporting process


bensternthal:
As part of CHIPS Alliance onboarding the project must adopt a security vulnerability reporting process.  You may also adopt the CHIPs policy:

https://github.com/chipsalliance/tac/blob/main/README.md#reporting-security-vulnerabilities

cunninghamsandwich:
I've filed a PR at the CHIPS Alliance org level to address this:  https://github.com/chipsalliance/.github/pull/5

This will enable a default process for all projects, which is to simply email security@lists.chipsalliance.org, as outlined in our TAC documentation:  https://github.com/chipsalliance/tac?tab=readme-ov-file#reporting-security-vulnerabilities

You're free to replace this with your own policy by creating a SECURITY.md file inside this repo at any time, but at least this will unblock for now.

FerralCoder:
I am working on the policy based on what we put in place for Caliptra. I will have it submitted by tomorrow at the latest.

PR #404, opened 2026-08-07
Use device config as orchestrator chain source

### LINKED ISSUE #375: One source of truth for the device list, with a build-time check
**What:** Have the device list and the orchestrator's list come from a single
shared source, or — if they stay as two lists — add a build-time check that
proves the two are exact mirrors: same number of devices, same order, same ids.
A shared id type is the cleanest way to make that check meaningful.

**Why:** There are two lists of devices today, and nothing makes sure they agree.
fwmanager just added one: the `MANAGED_DEVICES` list in
`target/mock/devices.rs`, built on the new `config.rs` schema. Its comment says
the order the devices are written in *is* the boot order — first one boots first,
and so on. The orchestrator sm has
its own list too: a `Chain` of `ComponentId`s, with its own order that also means
boot order. So both sides are independently deciding "who is device #1, who is
device #2, and what order they boot in."

The trap is that these two lists can quietly disagree. A device is identified
today only by where it sits in the list plus its `name` and `reset_line` — there
is no shared id. The fwmanager `validate()` checks that names aren't empty and
timeouts aren't zero, but it never checks that the list matches the
orchestrator's list. So if the two fall out of step — a different count, a
different order, a different device — nothing complains. It just boots or resets
the wrong device.

Since fwmanager owns the device list (the `config.rs` schema and the board
`devices.rs` tables), the shared id and the check belong on the fwmanager side —
shipping from `fwmanager-api` (or the board source it defines), with the
orchestrator's list reading those ids. The ideal is a single device source both
sides read from, so there's only ever one list; failing that, a build-time check
(a `const` assert) that fails the build when the two lists don't line up.

Daniiiil1:
Closes #375

## Summary
- move ComponentId into the dependency-free orchestrator API so board configuration and the state machine share one id type
- store the stable component id in DeviceConfig and reject empty tables or duplicate ids during const validation
- add Chain::from_device_configs, deriving chain size, order, and ids from the board array while looking up policy by id
- convert the mock board table to a fixed-size array with explicit ids

This removes the independently maintained positional-id/order convention. A board integration can no longer reorder its orchestrator chain separately from its device table when it uses the new constructor; the array length also fixes the Chain capacity at compile time.

## Validation
- bazel test //services/orchestrator/api:orchestrator_api_test
- bazel test //services/orchestrator/sm:orchestrator_sm_test
- bazel build //target/mock:devices
- rustfmt --check on all changed Rust files
- git diff --check

### DIFF
--- a/services/orchestrator/config/src/lib.rs
+++ b/services/orchestrator/config/src/lib.rs
@@ -7,6 +7,24 @@
 
 #![cfg_attr(not(test), no_std)]
 
+/// Stable identifier shared by the board device table and orchestrator core.
+///
+/// The value is opaque to the state machine. Keeping the type in this
+/// dependency-free config crate lets every layer refer to the same device without
+/// maintaining a second positional-id convention.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct ComponentId(u8);
+
+impl ComponentId {
+    pub const fn new(id: u8) -> Self {
+        Self(id)
+    }
+
+    pub const fn get(self) -> u8 {
+        self.0
+    }
+}
+
 /// What the orchestrator requires before it commits a staged image.
 ///
 /// Intentionally exhaustive (not `#[non_exhaustive]`): adding a variant is
@@ -67,6 +85,8 @@ pub struct BootCheckpoint<G> {
 /// Adding a field is a breaking change that updates every board table.
 #[derive(Debug, Clone, Copy)]
 pub struct DeviceConfig<R, G: 'static> {
+    /// Stable id used by the orchestrator, update manager, and board drivers.
+    pub id: ComponentId,
     pub name: &'static str,
     /// Reset signal id, passed to HalBootControl::new.
     pub reset_signal: R,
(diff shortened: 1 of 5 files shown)


chrysh:
I was working on the same problem in 9elements#26, with single source of truth as one important aspect: every fact lives in exactly one place, and anything derivable is not stored.

This PR branches from main and predates the #397 review round, so it reintroduces things that were already reviewed out there: `CommitPolicy` (redundant with kind — commit gating follows from Active/Passive), the free `validate()` (replaced by `const fn` constructors with private fields, so an invalid entry is unconstructible), and `BootSignal` in the schema (now a board-defined generic). Could you rework this on top of #397 instead? Until it merges, that's the `add-boot-walk` branch on the 9elements fork.

For the shape, take whatever is useful from 9elements#26 — it's a draft sketch of the same idea:

- kind, failure policy and `depends_on` (by device name) are declared in `DeviceConfig` itself, so no `attrs_for` mapping outside the table that could drift — and the closure being infallible means it can't even reject an unknown id.
- id is derived from table position instead of stored — declaration order is the boot order, and nothing persists ids across boots, so a stored id only creates a second ordering that can disagree (your `chain_builds_from_device_config_source_of_truth` test shows exactly that: `C1, C0` against table order).
- validation happens once, in a const `DeviceTable::new`; the chain conversion is infallible because a table that compiled is the proof, so `ChainError` goes away.

One thing from this PR that #26 should adopt rather than the other way around: the fixed-size array. `&[DeviceConfig; N]` carries the table size in the type and removes the last runtime assert in the conversion.

chrysh (changes requested):
Requesting changes per my comment above: please rework this on top of #397 — as-is it reintroduces `CommitPolicy`, the free `validate()`, and `BootSignal` in the schema, all of which were reviewed out there. Details and a reference sketch in the comment.

chrysh:
@rusty1968 this overlaps with the #397 outcomes and your CommitPolicy point; relevant to #407's single-source direction too.

PR #407, opened 2026-08-07
orchestrator-sm: carry recovery attempt on RecoverComponent

rusty1968:
`Effect::RecoverComponent` now carries `{ id, attempt }` instead of just `id`. `attempt` is the component's consecutive-recovery count (0 on the first try), taken straight from the core's retry counter.

Passing it in-band lets the Platform driver pick a recovery source per attempt (e.g. slot A → slot B → golden) without keeping its own counter, which could drift from the core's since the driver never sees when a recovery succeeds.

No behavior change in the core: the retry cap is still measured against the same counter; `attempt` only rides out on the effect. Tests updated.

### DIFF
--- a/services/orchestrator/sm/src/lib.rs
+++ b/services/orchestrator/sm/src/lib.rs
@@ -3,13 +3,10 @@
 
 //! `openprot_orchestrator_sm` — the eRoT boot-sequence state machine.
 //!
-//! This is the pure-reducer core ported from `rot_reducer`. It describes side
-//! effects as [`Effect`] values rather than performing them; the surrounding
-//! OpenPRoT shell carries them out via a [`Platform`] impl. No concrete hardware
-//! appears here — the machine is generic over an opaque [`ComponentId`].
-//!
-//! See `docs/verification-model.md` and `docs/state-machine.md` in the
-//! `rot_reducer` workspace for the full domain context and design rationale.
+//! This is the pure decision core: it describes side effects as [`Effect`]
+//! values rather than performing them; the surrounding OpenPRoT shell carries
+//! them out via a [`Platform`] impl. No concrete hardware appears here — the
+//! machine is generic over an opaque [`ComponentId`].
 //!
 //! Three invariants define the boundary:
 //!   1. **Effects flow through [`Sink`]** — fresh per event, drained afterward.
@@ -103,7 +100,7 @@ impl<const E: usize> Sink<E> {
     /// lockdown anyway, which is the louder signal.
     pub fn emit(&mut self, effect: Effect) {
         // Dead Err arm: overflow is proved impossible by `Rot::EFFECT_CAP_OK`
-        // (`E >= 2 * N + 2`) plus the reducer never emitting more than
+        // (`E >= 2 * N + 2`) plus the state machine never emitting more than
         // `2 * N + 2` effects into one Sink.
         let _ = self.effects.push(effect);
     }
@@ -473,7 +470,7 @@ impl<const N: usize, const E: usize> Rot<N, E> {
     }
 }
 
-/// The reducer proper: the per-state handlers, the superstate handler, and the
+/// The state machine proper: the per-state handlers, the superstate handler, and the
 /// entry actions. These are pure functions of `(stored data, state, event)` —
 /// they mutate [`Rot`]'s storage and push [`Effect`]s into the [`Sink`], and
 /// return an [`Outcome`] describing what should happen to the current state.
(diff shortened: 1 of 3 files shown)


services/orchestrator/sm/src/lib.rs
```diff
@@ -844,10 +841,20 @@ impl<const N: usize, const E: usize> Rot<N, E> {
...
-                ctx.emit(Effect::RecoverComponent(failed));
+                // `retry` here is the consecutive-recovery count for `failed`
+                // (0 on the first attempt); hand it to the driver so it need not
+                // track its own. It is bumped later, on `Restored`.
+                let attempt = match self.status_index(failed) {
+                    Some(i) => {
+                        self.statuses[i].released = false;
+                        self.statuses[i].retry
+                    }
+                    None => 0,
```
chrysh, line 852: Nit: Maybe a comment that this arm should be unreachable? Recovering(failed) can only be entered with an id that passed the dispatch-level chain-membership check, so status_index is always Some there.

chrysh:
Looks good — one gap: every assertion in the suite pins `attempt: 0`, so the counter actually incrementing is never covered. If `attempt` were hardcoded to 0, all tests would still pass.

Two tests that close that, verified green on this branch — feel free to take them as-is (they slot in after `corruption_while_recovering_retargets_to_new_component`):

```rust
/// `RecoverComponent.attempt` mirrors the core's retry counter: the second
/// consecutive recovery episode for the same component rides out with
/// `attempt: 1`, so a driver can escalate its recovery source without
/// keeping a counter of its own.
#[test]
fn recover_component_attempt_increments_across_episodes() {
    let (effects, _) = drive(
        passive_required(&[C0]),
        &[
            BOOT,
            Event::VerificationFailed(C0), // episode 1 → attempt 0
            Event::Restored(C0),           // credited: retry becomes 1, re-walk
            Event::VerificationFailed(C0), // episode 2 → attempt 1
        ],
    );
    assert!(effects.contains(&Effect::RecoverComponent { id: C0, attempt: 1 }));
}

/// A recovery episode displaced by another component's corruption ends
/// without a `Restored`, so it is never credited to the retry counter: the
/// displaced component's next episode reports `attempt: 0` again. Pins the
/// consecutive-count semantics — a driver must not assume attempts are
/// monotonic across displaced episodes.
#[test]
fn displaced_recovery_episode_restarts_attempt_count() {
    let (effects, state) = drive(
        passive_required(&[C0, C1]),
        &[
            BOOT,
            Event::VerificationFailed(C0), // → Recovering(C0), never credited
            Event::CorruptionDetected(C1), // displaces → Recovering(C1)
            Event::Restored(C1),           // C1 credited, re-walk from the top
            Event::VerificationFailed(C0), // C0's next episode…
        ],
    );
    // …did start (this assert is what makes the negative one meaningful)…
    assert_eq!(state, State::Recovering(C0));
    // …and is attempt 0 again, not 1.
    assert!(!effects.contains(&Effect::RecoverComponent { id: C0, attempt: 1 }));
}
```

The second one is worth a sentence in the `attempt` doc too: a displaced episode is never credited, so attempts are not monotonic across episodes.

services/orchestrator/sm/src/model.rs
```diff
@@ -304,9 +304,20 @@ pub enum Effect {
...
     /// recover; it does not encode how recovery is performed.
-    RecoverComponent(ComponentId),
+    ///
+    /// `attempt` is this component's consecutive-recovery count (0 on the first
+    /// attempt), taken straight from the core's own retry counter — the same
+    /// value the retry cap is measured against. It rides on the effect so the
+    /// driver can try a different recovery source each time (say, slot A on
+    /// attempt 0, slot B on 1, golden on 2) without counting attempts itself —
+    /// a count of its own could drift from the core's, since the driver never
+    /// sees when a recovery succeeds.
```
chrysh, line 316: Maybe add?: 

    /// Consecutive, not monotonic: a displaced episode is never credited,
    /// so that component's next episode reports `attempt: 0` again. Don't
    /// carry ladder progress across episodes.

chrysh (approved):
Tests and a doc suggestion for #407 are posted — take them as-is if they fit. From my side, merge #407 and #397 whenever you're happy, either order; they don't overlap.

rusty1968:
> Looks good — one gap: every assertion in the suite pins `attempt: 0`, so the counter actually incrementing is never covered. If `attempt` were hardcoded to 0, all tests would still pass.
> 
> Two tests that close that, verified green on this branch — feel free to take them as-is (they slot in after `corruption_while_recovering_retargets_to_new_component`):
> 
> ```rust
> /// `RecoverComponent.attempt` mirrors the core's retry counter: the second
> /// consecutive recovery episode for the same component rides out with
> /// `attempt: 1`, so a driver can escalate its recovery source without
> /// keeping a counter of its own.
> #[test]
> fn recover_component_attempt_increments_across_episodes() {
>     let (effects, _) = drive(
>         passive_required(&[C0]),
>         &[
>             BOOT,
>             Event::VerificationFailed(C0), // episode 1 → attempt 0
>             Event::Restored(C0),           // credited: retry becomes 1, re-walk
>             Event::VerificationFailed(C0), // episode 2 → attempt 1
>         ],
>     );
>     assert!(effects.contains(&Effect::RecoverComponent { id: C0, attempt: 1 }));
> }
> 
> /// A recovery episode displaced by another component's corruption ends
> /// without a `Restored`, so it is never credited to the retry counter: the
> /// displaced component's next episode reports `attempt: 0` again. Pins the
> /// consecutive-count semantics — a driver must not assume attempts are
> /// monotonic across displaced episodes.
> #[test]
> fn displaced_recovery_episode_restarts_attempt_count() {
>     let (effects, state) = drive(
>         passive_required(&[C0, C1]),
>         &[
>             BOOT,
>             Event::VerificationFailed(C0), // → Recovering(C0), never credited
>             Event::CorruptionDetected(C1), // displaces → Recovering(C1)
>             Event::Restored(C1),           // C1 credited, re-walk from the top
>             Event::VerificationFailed(C0), // C0's next episode…
>         ],
>     );
>     // …did start (this assert is what makes the negative one meaningful)…
>     assert_eq!(state, State::Recovering(C0));
>     // …and is attempt 0 again, not 1.
>     assert!(!effects.contains(&Effect::RecoverComponent { id: C0, attempt: 1 }));
> }
> ```
> 
> The second one is worth a sentence in the `attempt` doc too: a displaced episode is never credited, so attempts are not monotonic across episodes.

Thanks, I will make this an issue.

PR #408, opened 2026-08-08
hal: Add OTP capability traits

chrysh:
Capability traits for word-addressed one-time-programmable (OTP) memory.

## What

`hal/blocking/otp`: `OtpRead` (`words`, `read_word`) and `OtpProgram: OtpRead` (`program_word`), with `OtpErrorKind`/`OtpError` so generic code can react to failures without knowing the hardware. Dependency-free leaf package, following the flash package precedent.

## Design

- **One trait object per OTP region** — whatever the hardware partitions its fuse array into (AST10x0 data/config/SCU-protection regions, earlgrey `otp_ctrl` partitions).
- **Read and program are separate capabilities**, enforced at compile time: a consumer bound on `OtpRead` cannot program.
- **32-bit word addressing regardless of physical program granule.** Hardware with a wider granule implements `program_word` by re-programming the rest of the granule with its current value — a no-op for one-way bits.
- **`program_word` verifies by read-back.** A value that would reverse programmed bits fails with `AlreadyProgrammed` and leaves the word unchanged; re-programming the current value succeeds.
- **Out of scope:** what the words mean (keys, straps, SVN counters — service-level policy) and sub-word programming mechanics (soak/retry timing, redundancy voting — driver-internal).

A `MockOtp` test double with real one-time semantics follows in a separate PR.

https://claude.ai/code/session_01WfYp4c9FrMC5qHmwUk6FAK


leongross:
@chrysh, the Claude session link you posted is inaccessible for me, please double check acess permissions.

hal/blocking/otp/otp.rs
```diff
@@ -0,0 +1,92 @@
...
+impl<T: OtpRead + ?Sized> OtpRead for &mut T {
+    #[inline(always)]
+    fn words(&self) -> usize {
+        (**self).words()
+    }
+    #[inline(always)]
+    fn read_word(&mut self, index: usize) -> Result<u32, Self::Error> {
+        (**self).read_word(index)
+    }
+}
```
leongross, line 74: For AST10x0 and maybe even for Earl Grey, do we need to enforce read address alignment here? What is this for AST10x0, 32-bit?

leongross:
> Also, A MockOtp test double with real one-time semantics follows in a separate PR.

Can we get the tests in this PR as well? Then we can coherently check whether the default blanket implementations work as expected. The changes are quite small, so testing would, imho, not overload this PR.

chrysh:
There is already somewhere a traits repository and otp trait will be continued to be developed by @lxuxx

PR #411, opened 2026-08-11
earlgrey: Add SPI Device driver and SFDP support

anthonychen1251:
This PR adds the OpenTitan SPI Device (`spi_device`) driver into OpenPRoT, adds JESD216 SFDP table generation support, integrates the `spidev` service process into the Earlgrey HWE firmware, and adds an end-to-end (E2E) integration test on FPGA using `opentitanlib`.

### DIFF
--- a/target/earlgrey/drivers/spi_device.rs
+++ b/target/earlgrey/drivers/spi_device.rs
@@ -1,0 +1,680 @@
+// Licensed under the Apache-2.0 license
+// SPDX-License-Identifier: Apache-2.0
+
+//! OpenTitan SPI Device driver for Earlgrey.
+
+#![no_std]
+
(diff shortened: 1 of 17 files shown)


target/earlgrey/drivers/spi_device.rs
```diff
@@ -0,0 +1,680 @@
...
+        upload: true,
+        busy: true,
+        payload_io: None,
+        addr_mode: None,
+        dummy_cyc: 0,
+        filter: true,
+    };
+
+    pub const CHIP_ERASE2: Self = Self {
+        opcode: SpiFlashOpcode::CHIP_ERASE,
```
lchiawei: Not sure about it. Should we use CHIP_ERASE2 instead?

PR #418, opened 2026-08-14
orchestrator: Add the platform-driver skeleton behind the Platform seam

chrysh:
TL;DR: Adds `services/orchestrator/driver`, the effect-executing layer around the orchestrator state machine — the *platform driver* (#357's name for it; the last commit renames it from the working name "shell"). Only a skeleton: the actual implementation lands step by step.

- `board.rs` (trait-only commit): `ImageSource` — board-supplied access to a component's active firmware image (interposed flash, PLDM transfer, test double); `Verifier` — judges an image, where an error means the check could not run (a bad image is `Verdict::Rejected`, so an actuation fault cannot forge a verdict); `BoardTypes`/`Board` — one marker type and one bundle struct per board, so a new seam adds an associated type and a field, never another parameter.
- `driver.rs`: `PlatformDriver`, one executor method per SM effect, each documenting its obligation from the platform-boundary contract (orchestrator-model.md §6). `read_firmware`/`verify_firmware` are implemented; the rest return `NotImplemented` and the SM fail-closes on them.
- Executor-produced events queue in the driver and feed back via `PlatformDriver::take_event`. #419 proposes letting `Platform::execute` return the event instead, which deletes this queue.

The platform driver is board-agnostic: mock vs hardware is invisible from inside. Tests drive the real SM against a RAM-backed board.

### DIFF
--- a/services/orchestrator/driver/src/board.rs
+++ b/services/orchestrator/driver/src/board.rs
@@ -1,7 +1,7 @@
 // Licensed under the Apache-2.0 license
 // SPDX-License-Identifier: Apache-2.0
 
-//! What the board supplies to the shell: traits and wiring data only.
+//! What the board supplies to the driver: traits and wiring data only.
 //! Boards (or test mocks) implement these.
 
 use openprot_orchestrator_sm::ComponentId;
@@ -97,7 +97,7 @@ pub trait BoardTypes {
 }
 
 /// Everything the board supplies, built once at bring-up and handed to
-/// `Shell::new`. Fields are public: executors may need two parts at once
+/// `PlatformDriver::new`. Fields are public: executors may need two parts at once
 /// (disjoint borrows).
 ///
 /// ```ignore
--- a/services/orchestrator/driver/src/board.rs
+++ b/services/orchestrator/driver/src/board.rs
@@ -5,6 +5,7 @@
 //! Boards (or test mocks) implement these.
 
 use openprot_orchestrator_sm::ComponentId;
+use orchestrator_capabilities::BootControl;
 
 /// Access to one component's active firmware image, however it is reached —
 /// interposed flash, a PLDM/MCTP transfer, a RAM copy in tests.
@@ -93,8 +94,9 @@ pub trait BoardCapabilities {
     type Image: ImageSource;
     /// Judges images for every component.
     type Verifier: Verifier;
-    // Later seams: Reset (release/assert_reset), Evidence (checkpoint
-    // walk), Recovery, Staging.
+    /// Reset actuation for the managed components.
+    type BootControl: BootControl;
+    // Later seams: Evidence (checkpoint walk), Recovery, Staging.
 }
 
 /// Everything the board supplies, built once at bring-up and handed to
@@ -104,12 +106,14 @@ pub trait BoardCapabilities {
 /// ```ignore
 /// struct Ast1060Board;
 /// impl BoardCapabilities for Ast1060Board {
-///     type Image = SpiFlashImage;       // interposed flash, offsets from the slot layout
-///     type Verifier = ManifestVerifier; // signature + SVN via the crypto engine
+///     type Image = SpiFlashImage;         // interposed flash, offsets from the slot layout
+///     type Verifier = ManifestVerifier;   // signature + SVN via the crypto engine
+///     type BootControl = ExtrstGpio;      // per-component reset line
 /// }
 /// let board = Board::<Ast1060Board, 2> {
 ///     images: [bmc_image, cpld_image],
 ///     verifier,
+///     boot_controls: [bmc_reset, cpld_reset],
 /// };
 /// ```
 pub struct Board<B: BoardCapabilities, const N: usize> {
@@ -118,5 +122,8 @@ pub struct Board<B: BoardCapabilities, const N: usize> {
     pub images: [B::Image; N],
     /// Judges images for every component.
     pub verifier: B::Verifier,
-    // Later seams add fields, e.g. resets: [B::Reset; N].
+    /// `boot_controls[i]` actuates `ComponentId(i)`'s reset, same indexing
+    /// as `images`.
+    pub boot_controls: [B::BootControl; N],
+    // Later seams add fields, e.g. evidence: [B::Evidence; N].
 }
--- a/services/orchestrator/driver/src/board.rs
+++ b/services/orchestrator/driver/src/board.rs
@@ -17,8 +17,10 @@ pub trait ImageSource {
     /// Idempotent; a later `open` re-stages the image.
     fn open(&mut self) -> Result<(), Self::Error>;
 
-    /// Image length in bytes.
-    fn size(&mut self) -> Result<usize, Self::Error>;
+    /// Image length in bytes. Only meaningful after a successful `open`;
+    /// sources that learn the size during `open` cache it and report the
+    /// cached value here.
+    fn size(&self) -> Result<usize, Self::Error>;
 
     /// Reads `buf.len()` bytes starting at byte `offset` of the image.
     fn read_at(&mut self, offset: usize, buf: &mut [u8]) -> Result<(), Self::Error>;
@@ -33,7 +35,7 @@ impl<S: ImageSource> ImageSource for &mut S {
     }
 
     #[inline(always)]
-    fn size(&mut self) -> Result<usize, Self::Error> {
+    fn size(&self) -> Result<usize, Self::Error> {
         (**self).size()
     }
 
--- a/services/orchestrator/driver/src/board.rs
+++ b/services/orchestrator/driver/src/board.rs
@@ -80,7 +80,7 @@ impl<V: Verifier> Verifier for &mut V {
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
 pub enum Verdict {
     /// Reported as `Event::VerificationPassed`.
-    Authentic,
+    Authenticated,
     /// Reported as `Event::VerificationFailed`.
     Rejected,
 }
--- a/services/orchestrator/driver/src/board.rs
+++ b/services/orchestrator/driver/src/board.rs
@@ -85,9 +85,10 @@ pub enum Verdict {
     Rejected,
 }
 
-/// One board's type choices, named by a marker type. A new seam adds an
-/// associated type here and a field on [`Board`] — never another parameter.
-pub trait BoardTypes {
+/// The set of platform capabilities one board composes into the
+/// `PlatformDriver`, named by a marker type. A new seam adds an associated
+/// type here and a field on [`Board`] — never another parameter.
+pub trait BoardCapabilities {
     /// Image access for the managed components.
     type Image: ImageSource;
     /// Judges images for every component.
@@ -102,7 +103,7 @@ pub trait BoardTypes {
 ///
 /// ```ignore
 /// struct Ast1060Board;
-/// impl BoardTypes for Ast1060Board {
+/// impl BoardCapabilities for Ast1060Board {
 ///     type Image = SpiFlashImage;       // interposed flash, offsets from the slot layout
 ///     type Verifier = ManifestVerifier; // signature + SVN via the crypto engine
 /// }
@@ -111,7 +112,7 @@ pub trait BoardTypes {
 ///     verifier,
 /// };
 /// ```
-pub struct Board<B: BoardTypes, const N: usize> {
+pub struct Board<B: BoardCapabilities, const N: usize> {
     /// `images[i]` belongs to `ComponentId(i)` — device index = chain
     /// position = table declaration order.
     pub images: [B::Image; N],
--- a/services/orchestrator/driver/src/driver.rs
+++ b/services/orchestrator/driver/src/driver.rs
@@ -111,7 +111,7 @@ impl<B: BoardTypes, const N: usize> PlatformDriver<B, N> {
             .verify(id, source)
             .map_err(|_| DriverError::VerifierFault)?;
         self.enqueue(match verdict {
-            Verdict::Authentic => Event::VerificationPassed(id),
+            Verdict::Authenticated => Event::VerificationPassed(id),
             Verdict::Rejected => Event::VerificationFailed(id),
         })
     }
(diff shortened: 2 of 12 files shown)


leongross:
In https://github.com/OpenPRoT/openprot/pull/357 we discussed not naming a component shell due to the ambiguity of this name.

services/orchestrator/driver/src/board.rs
```diff
@@ -0,0 +1,121 @@
...
+    ) -> Result<Verdict, Self::Error> {
+        (**self).verify(id, image)
+    }
+}
+
+/// A [`Verifier`]'s judgment of one image.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Verdict {
+    /// Reported as `Event::VerificationPassed`.
+    Authentic,
```
leongross: nit: if we cryptographically authenticated an image, I would name it `Authenticated`, since this is common naming in cryptographic contexts.

services/orchestrator/driver/src/driver.rs
```diff
@@ -0,0 +1,233 @@
...
+        self.staged = Some(id);
+        Ok(())
+    }
+
+    /// Judge the staged image via the [`Verifier`] and queue the verdict:
+    /// `Event::VerificationPassed(id)` or `Event::VerificationFailed(id)`.
+    pub fn verify_firmware(&mut self, id: ComponentId) -> Result<(), ShellError> {
+        if self.staged != Some(id) {
+            return Err(ShellError::NoImage);
+        }
```
leongross, line 84: `NoImage` seems confusing, since this code location should inform the API consumer that the image is not staged and hence cannot be read, not that there is no image in the system for the Id, right? 

Maybe use `NotStaged` or sth similar.

services/orchestrator/shell/src/shell.rs
```diff
@@ -0,0 +1,233 @@
...
+    /// `Event::VerificationPassed(id)` or `Event::VerificationFailed(id)`.
+    pub fn verify_firmware(&mut self, id: ComponentId) -> Result<(), ShellError> {
+        if self.staged != Some(id) {
+            return Err(ShellError::NoImage);
+        }
+        let source = self
+            .board
+            .images
+            .get_mut(id.get() as usize)
+            .ok_or(ShellError::UnknownComponent)?;
```
leongross: Shouldn't the id validity check happen before we check if it is staged? Also, if the ID does not exist for the board images, it could never have been staged, right?

services/orchestrator/driver/src/tests.rs
```diff
@@ -0,0 +1,293 @@
...
+const IMAGE_MAGIC: [u8; 4] = *b"OPRT";
+const IMAGE_LEN: usize = 16;
+
+fn valid_image() -> std::vec::Vec<u8> {
+    let mut image = std::vec![0u8; IMAGE_LEN];
+    image[..4].copy_from_slice(&IMAGE_MAGIC);
+    image[4..IMAGE_LEN - 1].fill(0xAB);
+    image[IMAGE_LEN - 1] = image[..IMAGE_LEN - 1].iter().fold(0, |acc, b| acc ^ b);
+    image
+}
```
leongross, line 25: Will we at some point verify this exact image format? Then it would make sense to create an abstraction struct for this and serialze/deserialze it in a generic way and implement verification methods for this.
I agree it is overkill for a unit test, but when we have the image format at some point, we should remove this; it looks messy, and raw byte generation is prone to later image change regressions.

services/orchestrator/shell/src/shell.rs
```diff
@@ -0,0 +1,233 @@
...
+
+    fn enqueue(&mut self, event: Event) -> Result<(), ShellError> {
+        self.pending
+            .push_back(event)
+            .map_err(|_| ShellError::QueueFull)
+    }
+
+    /// Stage `id`'s image: open its source so
+    /// [`verify_firmware`](Self::verify_firmware) can read it.
+    pub fn read_firmware(&mut self, id: ComponentId) -> Result<(), ShellError> {
```
embediver: Should this be called `stage_firmware` or something similar instead?

services/orchestrator/driver/src/driver.rs
```diff
@@ -0,0 +1,233 @@
...
+impl core::error::Error for ShellError {}
+
+/// The effect executors. Everything device-specific lives in the [`Board`];
+/// the shell's own fields are bookkeeping.
+pub struct Shell<B: BoardTypes, const N: usize> {
+    board: Board<B, N>,
+    /// Component whose image is staged (source opened) for verification.
+    staged: Option<ComponentId>,
+    /// Events awaiting [`take_event`](Self::take_event).
+    pending: heapless::Deque<Event, EVENT_CAP>,
```
embediver: Without further explanation this seems a bit misplaced.
The shell does nothing beside storing events in a queue at this point while being responsible for executing effects otherwise.
Should the shell be more involved in handling events or should this be more separated?
(Coupling events stronger to the resulting effects?)

chrysh:
> In #357 we discussed not naming a component shell due to the ambiguity of this name.

True, I renamed it to PlatformDriver

services/orchestrator/shell/src/shell.rs
chrysh: Renamed

services/orchestrator/driver/src/driver.rs
chrysh: I want the orchestrator sm to handle the events, but wanted to discuss it with @rusty1968 in a separate PR: https://github.com/OpenPRoT/openprot/pull/419
The pending field is meant to be temporary only, until 419 is addressed and some version of that merged.

With that in mind, do you consider this PR merge-ready?

embediver: Will take a look at the other PR, probably tomorrow.

services/orchestrator/driver/src/driver.rs
```diff
@@ -0,0 +1,233 @@
...
+            Verdict::Rejected => Event::VerificationFailed(id),
+        })
+    }
+
+    /// Release `id` from reset, then walk its boot checkpoints; feed back
+    /// one `Event::ComponentReady(id)` (Active) or `Event::Booted(id)`
+    /// (Passive), or `Event::Timeout(id)` on window expiry.
+    pub fn release_reset(&mut self, _id: ComponentId) -> Result<(), DriverError> {
+        Err(DriverError::NotImplemented)
+    }
```
leongross: Instead of stubbing these functions, why not provide them by traits and extend the generic inherent implementation block over these traits? Something like:

```rust
type BootControl: orchestrator_capabilities::BootControl;
```

For the board, incorporate 

```rust
pub boot_controls: [B::BootControl; N]
```

And then have a generic implementation
```rust
pub fn assert_reset(&mut self, id: ComponentId) -> Result<(), DriverError> {
      self.board
          .boot_controls
          .get_mut(id.get() as usize)
          .ok_or(DriverError::UnknownComponent)?
          .hold_in_reset()
          .map_err(|_| DriverError::BootControlFault)
  }
```

services/orchestrator/driver/src/driver.rs
```diff
@@ -0,0 +1,233 @@
...
+        self.pending
+            .push_back(event)
+            .map_err(|_| DriverError::QueueFull)
+    }
+
+    /// Stage `id`'s image: open its source so
+    /// [`verify_firmware`](Self::verify_firmware) can read it.
+    pub fn stage_firmware(&mut self, id: ComponentId) -> Result<(), DriverError> {
+        self.staged = None;
+        let source = self
```
rusty1968: The index lookup appears twice. Here in `stage_firmware`  and in `verify_firmware`. We can extract a helper `source` so index arithmetic is performed in just one place.

```rust
fn source(&mut self, id: ComponentId) -> Result<&mut B::Image, DriverError> {
    self.board
        .images
        .get_mut(id.get() as usize)
        .ok_or(DriverError::UnknownComponent)
}
```

services/orchestrator/driver/src/board.rs
```diff
@@ -0,0 +1,121 @@
...
+    /// Reported as `Event::VerificationFailed`.
+    Rejected,
+}
+
+/// One board's type choices, named by a marker type. A new seam adds an
+/// associated type here and a field on [`Board`] — never another parameter.
+pub trait BoardTypes {
```
rusty1968: Pease rename `BoardTypes` to `BoardCapabilities`. This is an architectural clarification, not just a naming cleanup. The trait defines the set of platform capabilities that can be composed into a `PlatformDriver`,

services/orchestrator/driver/src/driver.rs
rusty1968: The stubs appear to undermine the composability described by the `BoardTypes` comment. 

- The comment says future seams such as Reset, Evidence, Recovery, and Staging will be added to the board type composition, but the PR currently represents them only as `PlatformDriver` methods returning `NotImplemented`. 

- Composing driver behavior by delegating to board-provided capabilities would let the implementation grow incrementally while preserving the intended separation between orchestration and platform-specific mechanisms.

rusty1968:
Once you address the comments about making the driver composable from the capabilities described in the BoardCapabilties trait - you should have the following code:

```rust 
impl<B: BoardCapabilities, const N: usize> PlatformDriver<B, N> {
    fn boot_control(
        &mut self,
        id: ComponentId,
    ) -> Result<&mut B::BootControl, DriverError> {
        self.board
            .boot_controls
            .get_mut(id.get() as usize)
            .ok_or(DriverError::UnknownComponent)
    }

    pub fn release_reset(&mut self, id: ComponentId) -> Result<(), DriverError> {
        self.boot_control(id)?
            .release()
            .map_err(|_| DriverError::BootControlFault)
    }

    pub fn assert_reset(&mut self, id: ComponentId) -> Result<(), DriverError> {
        self.boot_control(id)?
            .hold_in_reset()
            .map_err(|_| DriverError::BootControlFault)
    }
}
```

services/orchestrator/shell/src/shell.rs
rusty1968: I thought you had requested and I agreed for the word shell not to be part of the vocabulary -  `PlatformDriver` is a more accurate name than `Shell`: this component is the concrete platform-side implementation of the state machine's `Platform` seam. It receives `Effect` values and delegates them to composed board capabilities, rather than merely wrapping or placeholder-wiring the orchestrator.

services/orchestrator/driver/src/board.rs
chrysh: Renamed to `Authenticated` in da096c1.

services/orchestrator/driver/src/driver.rs
chrysh: Agreed, `NotStaged` says what the caller did wrong. Renamed in 1c1fea4.

services/orchestrator/shell/src/shell.rs
chrysh: Reordered in 7b54669: `verify_firmware` now does the id lookup first, so an unknown id reports `UnknownComponent` and only a known-but-not-staged id reports `NotStaged`. You're right that an unknown id can never have been staged — that's exactly why `NotStaged` was the misleading answer for it. Added a test for the unknown-id case.

services/orchestrator/driver/src/tests.rs
chrysh: Agreed — once a real image format exists, the byte poking gets replaced by a serialize/verify abstraction over that format. Until then the XOR image is a deliberate stand-in, so I'd leave it for this PR and swap it when the format lands.

services/orchestrator/driver/src/driver.rs
chrysh: Extracted in 49de7a4, with one deviation from your sketch: the helper takes the images array instead of `&mut self`, because in `verify_firmware` the returned image is used while `board.verifier` is borrowed — a `&mut self` helper would lock the whole struct and not compile there. Same single lookup site either way.

services/orchestrator/driver/src/board.rs
chrysh: Renamed in 4f7bbdb, and the trait doc now describes it as the set of platform capabilities composed into the `PlatformDriver`.

services/orchestrator/driver/src/driver.rs
chrysh: Done for the Reset seam in f3399a7: `BoardCapabilities` gains `type BootControl: orchestrator_capabilities::BootControl`, `Board` gains `boot_controls: [B::BootControl; N]`, and `release_reset`/`assert_reset` delegate to it via a `boot_control()` helper — matching your snippet. The remaining stubs stay `NotImplemented` because their capability traits don't exist yet: `BootWatch` is the next candidate (completes `release_reset`'s checkpoint feedback) but is poll-based with injected time, so wiring it touches the event-loop design being discussed in #419. The others (recovery, staging, SVN floor, attestation, requester reports) get composed the same way as their traits land.

rusty1968: Usage of stubs is incompatible with the design. If you are using capabilties and delegation there is no reason for doing stubbing in the concrete type.

chrysh: Removed the stubs. If all is fine now, can you approve and merge?

services/orchestrator/driver/src/driver.rs
chrysh: Platform::execute now returns the follow-up event, so verify_firmware hands its verdict back directly. The pending deque, take_event and QueueFull are gone.

services/orchestrator/driver/src/board.rs
```diff
@@ -0,0 +1,129 @@
...
+
+//! What the board supplies to the driver: traits and wiring data only.
+//! Boards (or test mocks) implement these.
+
+use openprot_orchestrator_sm::ComponentId;
+use orchestrator_capabilities::BootControl;
+
+/// Access to one component's active firmware image, however it is reached —
+/// interposed flash, a PLDM/MCTP transfer, a RAM copy in tests.
+pub trait ImageSource {
```
rusty1968, line 12: The trait appears to over-constrain implementers by requiring &mut self for operations that are conceptually read-only (size, and potentially read_at).

rusty1968 (approved):

services/orchestrator/driver/src/board.rs
chrysh: I agree with the size parameter. The other read functions in the repo take &mut self, so I would leave it as is.

leongross (approved):
lgtm

PR #419, opened 2026-08-14
orchestrator: Let Platform::execute return a follow-up event

chrysh:
TL;DR: Resolved per review: blocking `execute` is fine, and the queue lives inside the SM.

An executor that learns something synchronously — a verification verdict — has no way to feed it back through `Platform::execute`, so every shell needs its own event queue drained between dispatches.

`execute` now returns `Result<Option<Event>, EffectError>`: the SM queues the driver's returned event and settles it in the same run, so one dispatch runs to quiescence. The internal pending queue pops as it settles, so it bounds in-flight events instead of a run's total length; a queue overflow injects `EffectFailed` and latches `Locked` — the same fail-closed rule as a failed effect. At most one event per effect; results that arrive later (boot progress, timer expiry) stay outside events.

Breaking change to the `Platform` seam. The ast10x0 runtime QEMU test's `FakePlatform` is adapted in the same commit. The shell (#418) currently works around the gap with a shell-side queue and drops it once this lands.

### DIFF
--- a/services/orchestrator/sm/src/lib.rs
+++ b/services/orchestrator/sm/src/lib.rs
@@ -879,7 +879,7 @@ impl<const N: usize, const E: usize> Rot<N, E> {
 
 /// Signals that the platform driver could not carry out an [`Effect`]. The machine does
 /// not need the driver's error detail — **every** actuation failure is treated
-/// the same, fail-closed: the driver injects [`Event::EffectFailed`] and the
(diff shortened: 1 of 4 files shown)


rusty1968 (approved):

rusty1968:
It is fine to have a blocking execute and place the queue inside the SM.

rusty1968:
In the PR description : "the driver queues a returned event" should read "the SM queues the driver's returned event." Driver queues nothing.

PR #421, opened 2026-08-17
orchestrator: Test the at-rest guarantee end to end

chrysh:
Now that the platform driver (#418) has merged with the reset wiring already composed in (`BootControl` in `BoardCapabilities`, `assert_reset`/`release_reset` delegating to it), this PR is reduced to what 418 did not carry: the end-to-end proofs of the at-rest guarantee.

- `release_follows_verification`: the component is still held in reset while its image is verified (observed from inside the verification window) and the line is released only on the passing verdict.
- `failed_release_fails_closed`: a dead reset line is a failed actuation, not a verdict — the SM latches Locked and the component never leaves reset.
- `BootControl` is re-exported from the driver crate so board crates need only one import.

Part of 9elements/openprot#4.

### DIFF
--- a/services/orchestrator/driver/src/tests.rs
+++ b/services/orchestrator/driver/src/tests.rs
@@ -76,15 +76,15 @@ impl ImageSource for MemImage {
 }
 
 #[derive(Debug)]
-struct VerifierBroken;
+struct VerifierError;
 
-impl core::fmt::Display for VerifierBroken {
+impl core::fmt::Display for VerifierError {
     fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
         f.write_str("verifier broken")
     }
 }
 
-impl core::error::Error for VerifierBroken {}
+impl core::error::Error for VerifierError {}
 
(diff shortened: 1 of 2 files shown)


services/orchestrator/driver/src/tests.rs
```diff
@@ -388,3 +388,80 @@ fn execute_returns_the_verdict_event() {
...
+/// Wraps [`XorVerifier`] and snapshots the reset line as the check runs,
+/// so the test can see the line state inside the verification window.
+struct LineWatchingVerifier {
+    inner: XorVerifier,
+    line: std::rc::Rc<core::cell::Cell<bool>>,
+    held_during_verify: std::rc::Rc<core::cell::Cell<bool>>,
+}
+
+impl Verifier for LineWatchingVerifier {
+    type Error = VerifierBroken;
```
leongross: Must have slipped in the past review, please rename the VerifierBroken error struct to `VerifierError`.

leongross (changes requested):

leongross (approved):
Looks good, thanks!

PR #422, opened 2026-08-17
orchestrator: Add the SvnFloor anti-rollback capability trait

chrysh:
TL;DR: The software version number counter abstraction that sits on top of OTP, etc.

One managed device's durable SVN floor behind the capability seam, mirroring `BootControl`: `floor()` / `advance(to)`, `type Error: core::error::Error`, no HAL bounds.

The contract does the anti-rollback work:
- **Monotonic.** `advance` to a floor at or below the current one is a successful no-op — a replayed commit is harmless — and no call ever lowers the floor.
- **Durable on return.** `Ok` from `advance` means the new floor survives power loss; a torn write may lose an advance, never lower one.
- Storage encoding (OTP fuse counters, protected flash, mocks) stays behind the trait; PLDM firmware devices that commit internally simply have no eRoT-side floor.

Trait only. Follow-ups: the OTP-backed adapter (on the `hal_otp` traits) and the platform driver's `commit_svn_floor` executor.

### DIFF
--- a/services/orchestrator/capabilities/src/svn_floor.rs
+++ b/services/orchestrator/capabilities/src/svn_floor.rs
@@ -39,7 +39,7 @@ pub trait SvnFloor {
     type Error: core::error::Error;
 
     /// The current floor: the lowest SVN this device may still boot.
-    fn floor(&mut self) -> Result<Svn, Self::Error>;
+    fn floor(&self) -> Result<Svn, Self::Error>;
 
     /// Raises the floor to `to`. At or below the current floor: `Ok`, no-op.
     fn advance(&mut self, to: Svn) -> Result<(), Self::Error>;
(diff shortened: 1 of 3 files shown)


services/orchestrator/capabilities/src/svn_floor.rs
```diff
@@ -0,0 +1,130 @@
...
+pub trait SvnFloor {
+    /// The error type of this device's floor storage.
+    ///
+    /// Bounded by [`core::error::Error`] so the orchestrator gets `Display`
+    /// and a `source()` cause chain, not just a `Debug` dump. Error
+    /// categories are implementation-defined.
+    type Error: core::error::Error;
+
+    /// The current floor: the lowest SVN this device may still boot.
+    fn floor(&mut self) -> Result<Svn, Self::Error>;
```
rusty1968: Reading a value does not require exclusive access. 

```rust
fn floor(&self)  -> Result<Svn, Self::Error>;
```

rusty1968 (changes requested):
just that one feedback and we are good to go.

services/orchestrator/capabilities/src/svn_floor.rs
chrysh: Fixed

rusty1968 (approved):

PR #425, opened 2026-08-18
orchestrator: Device table is the single source of the chain of trust

chrysh:
Makes the board device table the single source of truth for the orchestrator's chain of trust, and gives every invariant exactly one checker:

- `ComponentKind`/`FailurePolicy` move into `orchestrator-config`; the SM re-exports them (API unchanged).
- `DeviceConfig` declares kind, failure policy, and `depends_on` (by device name); the new const `DeviceTable` wrapper is the only place cross-entry invariants are checked, so an invalid table fails the build.
- `Chain::from_table` derives the chain from the table (index = id, declaration order = walk order) and is infallible — holding a `DeviceTable` is the proof — so `ChainError` and the `TryFrom` impl are deleted.
- The mock board derives `DEVICE_COUNT`/`EFFECT_CAP` from its table, declares `MAX_RETRY`, and gains a host test driving table → chain → orchestrator end to end.

Draft: overlaps with #394 in the device-table schema (both add fields to `DeviceConfig`); whichever lands first, the other rebases.

### DIFF
--- a/services/orchestrator/config/src/lib.rs
+++ b/services/orchestrator/config/src/lib.rs
@@ -6,9 +6,13 @@
 //! device is named here.
 //!
 //! Invariants are enforced in the `const fn` constructors, so an invalid
-//! table is a build error and there is no validate step to forget. Checks
-//! on board-defined types belong next to the table that gives them
-//! meaning (`target/mock/devices.rs` shows the pattern).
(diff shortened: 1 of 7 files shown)


chrysh:
Closing in favor of #404, which covers the same ground.

PR #431, opened 2026-08-19
orchestrator: Add the Updatable update capability trait

chrysh:
Adds the `Updatable` capability trait to the orchestrator's capability crate, covering the staging and activation half of the update flow.

- `poll_stage` advances the transfer one bounded step per call, so a single-threaded runtime stays live and cancellation takes effect between steps.
- `activate` marks the staged image as the boot candidate only tentatively; committing the image is TrialBoot's job.
- `PayloadSource` is the chunked read seam, serving both direct-flash adapters and PLDM devices that pull their own chunks.

Trait-level tests included in `updatable.rs`.

Part of 9elements/openprot#8. Related: #377 — the trial-boot/commit/rollback capabilities; `activate` here stays tentative because committing is that seam's job.

### DIFF
--- a/services/orchestrator/capabilities/src/updatable.rs
+++ b/services/orchestrator/capabilities/src/updatable.rs
@@ -1,0 +1,730 @@
+// Licensed under the Apache-2.0 license
+// SPDX-License-Identifier: Apache-2.0
+
+//! The [`Updatable`] update capability contract.
+
+/// Update capability: stage a payload on one managed device and mark the
+/// staged image as its boot candidate.
+///
+/// The two operations: write to the inactive slot only, then update
+/// slot metadata to prefer the new image. Universal across device
+/// archetypes: a direct-flash adapter writes the payload into the
+/// device's inactive slot itself, a PLDM adapter drives the device's
+/// own transfer. Slot identity never crosses the
+/// seam — which slot is inactive is device state, so
+/// [`activate`](Self::activate) can only ever mean "the image the last
+/// completed staging delivered".
+///
+/// The caller is the platform driver. The state machine only decides
+/// that an update runs; the driver owns the adapter, polls the transfer
+/// on its own clock, and reports the outcome back as an event. The
+/// update source that delivered the payload fills what backs
+/// [`PayloadSource`] and never drives the device.
+///
+/// What this trait deliberately does not claim:
+///
+/// - **Verification** runs on the candidate before staging as
+///   orchestrator policy; post-write read-back is the optional
+///   `ReadBack` capability.
+/// - **Commit.** Activation is always tentative: it proposes the staged
+///   image as the preferred boot target, never commits it. The commit
+///   gate is the optional `TrialBoot` capability, which resolves what
+///   `activate` proposed; there is no second slot-selection owner.
+/// - **Booting.** Resetting the device into the candidate is
+///   [`BootControl`](crate::BootControl). When activation takes effect
+///   (next reset, or a device-internal restart on self-activating
+///   devices) is device-defined; sequencing belongs to the flows.
+///
+/// # Contract
+///
+/// - **Staging is inert.** However staging ends — fault, abandon, power
+///   loss — the active image is untouched: the staging area is inactive
+///   by construction. Staging anew is always allowed and discards any
+///   previously staged, unactivated payload.
+/// - **Staging is polled, never blocking.** A payload is tens of
+///   megabytes and a transfer takes minutes; each
+///   [`poll_stage`](Self::poll_stage) call does one step, at most one
+///   payload pull plus one device transaction, and returns without
+///   waiting on the device. A busy device is not an error: the step
+///   returns [`Transferring`](StageProgress::Transferring) with `written`
+///   unchanged. So a single-threaded runtime stays live, the update
+///   source gets progress, and abandoning mid-transfer is
+///   [`abandon`](Self::abandon) instead of waiting out a blocked call.
+///   Liveness policy stays with the caller: it watches `written` and
+///   abandons a transfer that stalls too long, on its own clock.
+/// - **`Ready` means ready.** The device holds the complete payload and
+///   `activate` may be called. `activate` in any other staging state is
+///   an error.
+pub trait Updatable {
+    /// The error type of this device's update path.
+    ///
+    /// Bounded by [`core::error::Error`] so the orchestrator gets
+    /// `Display` and a `source()` cause chain, not just a `Debug` dump.
+    /// Error categories are implementation-defined.
+    type Error: core::error::Error;
+
+    /// Advances staging by one step, pulling from `payload`.
+    ///
+    /// A step is bounded: at most one pull from `payload` and at most one
+    /// device transaction, never a wait for device progress. Returning
+    /// with `written` unchanged is legal (busy device, PLDM retransmit) and
+    /// is not an error.
+    ///
+    /// The first call from idle (fresh device, after [`Ready`], an error,
+    /// or [`abandon`](Self::abandon)) starts a new transfer; the caller
+    /// keeps polling with the same `payload` until [`Ready`] or an error.
+    /// The implementor pulls at whatever offsets its transfer needs (a
+    /// PLDM device requests its own chunks, including retransmits);
+    /// `payload` must serve any in-range read.
+    ///
+    /// Generic for static dispatch, which makes `Updatable` itself
+    /// non-dyn-compatible; the associated `Error` type effectively
+    /// already did.
+    ///
+    /// [`Ready`]: StageProgress::Ready
+    fn poll_stage(&mut self, payload: &impl PayloadSource) -> Result<StageProgress, Self::Error>;
+
+    /// Discards the in-progress transfer or staged, unactivated payload.
+    ///
+    /// Infallible: back to idle unconditionally. Cleanup a device needs
+    /// (marking a half-written slot dirty) is the implementor's, deferred
+    /// to the next staging if it must touch hardware.
+    fn abandon(&mut self);
+
+    /// Marks the staged image as the device's boot candidate
+    /// (tentative; see the trait docs on commit).
+    ///
+    /// [`Ready`] persists until a new staging starts or
+    /// [`abandon`](Self::abandon), and `activate` while [`Ready`] is
+    /// idempotent: a repeated call succeeds.
+    ///
+    /// [`Ready`]: StageProgress::Ready
+    fn activate(&mut self) -> Result<(), Self::Error>;
+}
+
+/// What one [`Updatable::poll_stage`] step established.
+///
+/// Intentionally exhaustive (not `#[non_exhaustive]`): adding a state is a
+/// breaking change, so every consumer handles it explicitly.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum StageProgress {
+    /// Transfer ongoing; poll again. `written`/`total` bytes feed the Update
+    /// Source's progress report.
+    Transferring {
+        /// Bytes written to the device so far. Monotonic and below `total`,
+        /// but free to hold still across calls (busy device,
+        /// retransmit); a caller deciding when a transfer has stalled
+        /// keys on this value.
+        written: u64,
+        /// Total payload bytes.
+        total: u64,
+    },
+    /// The device holds the complete payload; `activate` may be called.
+    Ready,
+}
+
+/// Chunked, random-access read seam [`Updatable::poll_stage`] pulls from.
+///
+/// The candidate payload is streamed and never RAM-resident;
+/// this is the window a device adapter reads it through. Where the bytes
+/// live — frontend staging flash, a mapped blob, a test slice — stays
+/// behind the source.
+pub trait PayloadSource {
+    /// Total payload length in bytes, constant for the lifetime of the
+    /// source: adapters allocate staging buffers from it and treat the
+    /// transfer as complete once this many bytes are written.
+    fn len(&self) -> u64;
+
+    /// True if the payload is empty.
+    fn is_empty(&self) -> bool {
+        self.len() == 0
+    }
+
+    /// Fills `buf` from `offset`. The read is exact: short fills are a
+    /// fault, and `offset + buf.len()` beyond [`len`](Self::len) is out
+    /// of range. There are no partial reads: the length is known up
+    /// front, so a short read can only mean the source cannot serve
+    /// what `len` promised, and a partial-read API would put a retry
+    /// loop into every adapter.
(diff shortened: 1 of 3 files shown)


leongross:
@chrysh I can't inspect your claude session links, please fix the permissions

chrysh:
> @chrysh I can't inspect your claude session links, please fix the permissions

Sorry, there wasn't meant to be a session link. The commit is still in draft state, because I didn't have the time to clean it up yet.

services/orchestrator/capabilities/src/updatable.rs
```diff
@@ -0,0 +1,356 @@
...
+    /// Advances staging by one bounded step, pulling from `payload`.
+    ///
+    /// The first call from idle (fresh device, after [`Ready`], an error,
+    /// or [`abandon`](Self::abandon)) starts a new transfer; the caller
+    /// keeps polling with the same `payload` until [`Ready`] or an error.
+    /// The implementor pulls at whatever offsets its transfer needs (a
+    /// PLDM device requests its own chunks, including retransmits);
+    /// `payload` must serve any in-range read.
+    ///
+    /// Generic for static dispatch; `?Sized` admits `&dyn PayloadSource`.
```
rusty1968: I'd prefer to keep the API statically dispatched and avoid introducing trait-object support unless we have a concrete use case.

services/orchestrator/capabilities/src/updatable.rs
```diff
@@ -0,0 +1,356 @@
...
+    /// This makes `Updatable` non-dyn-compatible, which the associated
+    /// `Error` type effectively already did.
+    ///
+    /// [`Ready`]: StageProgress::Ready
+    fn poll_stage(
```
rusty1968: What prevents an implementation from blocking indefinitely inside poll_stage()? The trait docs require a "bounded step", but no bound or progress guarantee is specified.

services/orchestrator/capabilities/src/updatable.rs
chrysh: Changed

services/orchestrator/capabilities/src/updatable.rs
chrysh: The doc was updated.

services/orchestrator/capabilities/src/updatable.rs
```diff
@@ -0,0 +1,677 @@
...
+    /// loop into every adapter.
+    fn read_at(&self, offset: u64, buf: &mut [u8]) -> Result<(), PayloadFault>;
+}
+
+/// Why a payload read failed — the one distinction retry policy needs.
+///
+/// No further detail crosses the seam (mirroring `BootWatch`): the source
+/// logs the concrete cause while it is still in scope.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum PayloadFault {
```
leongross: It's an Error; why not call it `PayloadError`

services/orchestrator/capabilities/src/updatable.rs
```diff
@@ -0,0 +1,677 @@
...
+
+        fn poll_stage(&mut self, payload: &impl PayloadSource) -> Result<StageProgress, MockFault> {
+            if self.ready {
+                self.abandon(); // a poll after Ready starts a new transfer
+            }
+            let total = usize::try_from(payload.len()).unwrap();
+            if self.done == 0 {
+                self.staged = vec![0; total];
+            }
+            let end = (self.done + 2).min(total);
```
leongross: How is this 2 tied to MockDevice properties? Looks like it could easily break once more tests accumulate.

services/orchestrator/capabilities/src/updatable.rs
```diff
@@ -0,0 +1,677 @@
...
+            if let Err(fault) = payload.read_at(self.done as u64, &mut self.staged[self.done..end])
+            {
+                self.abandon();
+                return Err(MockFault::Pull(fault));
+            }
+            self.done = end;
```
leongross: done is irritating wording. do you mean something like count could work?

services/orchestrator/capabilities/src/updatable.rs
```diff
@@ -0,0 +1,677 @@
...
+    fn stage_all<D: Updatable>(dev: &mut D, payload: &impl PayloadSource) -> Result<(), D::Error> {
+        loop {
+            if let StageProgress::Ready = dev.poll_stage(payload)? {
+                return Ok(());
+            }
+        }
+    }
+
+    #[test]
+    fn contract_is_implementable_without_the_hal() {
```
leongross: None of these tests use the HAL; why make it explicit here? confusing

services/orchestrator/capabilities/src/updatable.rs
```diff
@@ -0,0 +1,677 @@
...
+        let payload = SliceSource(b"image");
+
+        dev.poll_stage(&payload).expect("first step failed");
+        dev.abandon();
+
+        assert_eq!(dev.activate(), Err(MockFault::NothingStaged));
+        // A fresh transfer starts from byte zero.
+        assert_eq!(
+            dev.poll_stage(&payload),
+            Ok(StageProgress::Transferring { done: 2, total: 5 })
```
leongross: you are always using these hard-coded values, they are easy to drift away over time. You are re-using it, so define it once for the test and operate of the attributes lengths etc to keep it consistent.

services/orchestrator/capabilities/src/updatable.rs
```diff
@@ -0,0 +1,677 @@
...
+    // A direct-flash adapter, the erase-before-write shape: one step is
+    // one flash operation, either erasing the next sector or programming
+    // the next page. Erase steps pull nothing and hold `done` still.
+    const FLASH_PAGE: usize = 2;
+    const FLASH_SECTOR: usize = 4;
+
+    struct MockFlashDevice {
+        slot: [u8; 2 * FLASH_SECTOR],
+        erased: [bool; 2],
+        done: usize,
```
leongross: `read` or `counter`, not `done`

services/orchestrator/capabilities/src/updatable.rs
```diff
@@ -0,0 +1,677 @@
...
+
+    // A direct-flash adapter, the erase-before-write shape: one step is
+    // one flash operation, either erasing the next sector or programming
+    // the next page. Erase steps pull nothing and hold `done` still.
+    const FLASH_PAGE: usize = 2;
+    const FLASH_SECTOR: usize = 4;
+
+    struct MockFlashDevice {
+        slot: [u8; 2 * FLASH_SECTOR],
+        erased: [bool; 2],
```
leongross: What exactly is `FLASH_PAGE` used for? Isn't that what erased and slot ar based on with their `2`?

services/orchestrator/capabilities/src/updatable.rs
```diff
@@ -0,0 +1,677 @@
...
+            if self.ready {
+                self.abandon();
+            }
+            let total = usize::try_from(payload.len()).unwrap();
+            // An oversized payload is a device error, not a caller bug:
+            // rejected before anything is erased or written.
+            if total > self.slot.len() {
+                return Err(MockFault::ExceedsSlot);
+            }
+            let sector = self.done / FLASH_SECTOR;
```
leongross: again, the `done` wording confuses here imo

services/orchestrator/capabilities/src/updatable.rs
```diff
@@ -0,0 +1,677 @@
...
+
+    #[test]
+    fn a_flash_shaped_device_fits_the_seam() {
+        let mut dev = MockFlashDevice::idle();
+        let payload = SliceSource(b"8 bytes!");
+
+        // Erase sector 0: a step with no pull, `done` holds still.
+        assert_eq!(
+            dev.poll_stage(&payload),
+            Ok(StageProgress::Transferring { done: 0, total: 8 })
```
leongross: make this depend on the length of payload so it can't drift on change

services/orchestrator/capabilities/src/updatable.rs
chrysh: Renamed to `PayloadReadError` — scoped to the read, since `Updatable::Error` is the update path's error type.

services/orchestrator/capabilities/src/updatable.rs
chrysh: Defined `MOCK_STEP`; the asserts now derive from the consts and `payload.len()`.

services/orchestrator/capabilities/src/updatable.rs
chrysh: Renamed to `count`.

services/orchestrator/capabilities/src/updatable.rs
chrysh: Renamed to `contract_is_implementable`.

services/orchestrator/capabilities/src/updatable.rs
chrysh: Defined `MOCK_STEP`/`FLASH_SECTORS`; asserts now derive from the consts and `payload.len()`.

services/orchestrator/capabilities/src/updatable.rs
chrysh: Renamed to `count`.

services/orchestrator/capabilities/src/updatable.rs
chrysh: Page = program granularity, sector = erase granularity; the `2` in `slot`/`erased` was the sector count — now the `FLASH_SECTORS` const with inline comments.

services/orchestrator/capabilities/src/updatable.rs
chrysh: Renamed to `count`.

services/orchestrator/capabilities/src/updatable.rs
chrysh: The asserts now derive from `FLASH_PAGE`/`FLASH_SECTOR` and `payload.len()`, with a guard that the payload exactly fills the slot.

services/orchestrator/capabilities/src/updatable.rs
```diff
@@ -0,0 +1,710 @@
...
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum StageProgress {
+    /// Transfer ongoing; poll again. `done`/`total` bytes feed the Update
+    /// Source's progress report.
+    Transferring {
+        /// Bytes the device holds so far. Monotonic and below `total`,
+        /// but free to hold still across calls (busy device,
+        /// retransmit); a caller deciding when a transfer has stalled
+        /// keys on this value.
+        done: u64,
```
leongross: change to count or written

leongross (approved):

services/orchestrator/capabilities/src/updatable.rs
chrysh: Done renaming is done

services/orchestrator/capabilities/src/updatable.rs
chrysh: This is now the concrete use case: boards are usually heterogeneous (flash and PLDM devices side by side), and the generic `poll_stage` plus associated `Error` forced one concrete `Updatable` type per board, i.e. a hand-written delegation enum in every board crate. The second commit makes the trait dyn-compatible following the `BootWatch` pattern (erased `UpdateError`, detail logged in scope — `Display` and `source()` still there). Static dispatch stays available by wrapping in an enum, it's just no longer mandatory. See `a_heterogeneous_fleet_stages_through_one_seam`.

services/orchestrator/capabilities/src/updatable.rs
```diff
@@ -0,0 +1,748 @@
...
+/// The candidate payload is streamed and never RAM-resident;
+/// this is the window a device adapter reads it through. Where the bytes
+/// live — frontend staging flash, a mapped blob, a test slice — stays
+/// behind the source.
+pub trait PayloadSource {
+    /// Total payload length in bytes.
+    fn len(&self) -> u64;
+
+    /// True if the payload is empty.
+    fn is_empty(&self) -> bool {
```
rusty1968, line 183: What is supposed to happen when you stage an empty payload?

services/orchestrator/capabilities/src/updatable.rs
```diff
@@ -0,0 +1,748 @@
...
+                offset: 0,
+                retransmitted: false,
+                ready: false,
+                active: false,
+            }
+        }
+    }
+
+    impl Updatable for MockPldmDevice {
+        fn poll_stage(
```
rusty1968, line 508: Seems you can activate an empty payload?

let total = usize::try_from(payload.len()).unwrap();   // total = 0
if self.count == 0 {
    self.staged = vec![0; total];                      // vec![0; 0]  -> empty buffer
}
let end = (self.count + MOCK_STEP).min(total);         // (0 + 2).min(0) = 0
// read_at(0, &mut self.staged[0..0])  -> zero-length read, Ok
self.count = end;                                       // count = 0
if self.count == total {                               // 0 == 0  -> true
    self.ready = true;
    Ok(StageProgress::Ready)                            // returns Ready on the FIRST call
}

services/orchestrator/capabilities/src/updatable.rs
```diff
@@ -0,0 +1,748 @@
...
+
+/// Chunked, random-access read seam [`Updatable::poll_stage`] pulls from.
+///
+/// The candidate payload is streamed and never RAM-resident;
+/// this is the window a device adapter reads it through. Where the bytes
+/// live — frontend staging flash, a mapped blob, a test slice — stays
+/// behind the source.
+pub trait PayloadSource {
+    /// Total payload length in bytes.
+    fn len(&self) -> u64;
```
rusty1968, line 180: Is len() constant for the lifetime of the source? The Mock implementation already depends on it.

services/orchestrator/capabilities/src/updatable.rs
embediver: I would guess that staging an empty payload should be allowed (since there can be no assumption made about the validity, a 1 byte payload is not empty, but is likely not valid as well).

But I'm not sure if `is_empty` would be useful then.

chrysh: Why should it be allowed? It most probably is not what the caller wanted to do. I think I would rather return an error. I think 0 bytes are way more probable to occur in code than 1 byte?

services/orchestrator/capabilities/src/updatable.rs
chrysh: edited the comment

services/orchestrator/capabilities/src/updatable.rs
chrysh: Fixed and added test

PR #433, opened 2026-08-20
Ast10x0 : FMC flash backend for the flash service

rusty1968:
This PR depends on https://github.com/OpenPRoT/openprot/pull/365

## Tests

The service is exercised end-to-end through the full IPC stack (FlashIpcClient → server → BlockingFlash → Ast10x0FmcFlashDriver → FMC CS0), not just at the peripheral level.

- flash_qemu_test — runs under emulation, no hardware required
- flash_evb_test — the same system image on real AST10x0 EVB hardware

### DIFF
--- a/MODULE.bazel
+++ b/MODULE.bazel
@@ -22,7 +22,7 @@ local_path_override(
 
 git_override(
     module_name = "pigweed",
-    commit = "6f0aac30f8313cbd5103686097a1f6b3098e8aac",  # roll:pigweed
+    commit = "f540ae6c33424230ed2d388599efa89b6baacf80",
     patch_args = ["-p1"],
     patches = [
         # Program the VeeR external-interrupt redirect table (MEIVT) in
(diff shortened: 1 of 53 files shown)


yvettewu1:
Tested with  ast1060 evb board: 
1. The flash test itself passed: the client completed geometry, erase, program, read-back, and restore, then printed flash client PASS. 
2. The later HardFault was caused by the target shutdown path calling semihosting exit(), which works under QEMU/debugger-style environments but can trap as a HardFault on bare AST10x0 EVB hardware.

Tested with the following changes and no more HardFault:

-use cortex_m_semihosting::debug::{exit, EXIT_FAILURE, EXIT_SUCCESS};
+use console_backend::console_backend_write_all;

-exit(status);
+let _ = console_backend_write_all(sentinel);

rusty1968:
Great job!

PR #434, opened 2026-08-20
target/mock: Add MockOtpFloor, the mock board's OTP fuse bank

chrysh:
Testonly-free home with the mock board: `target/mock/otp.rs` (`board_otp`). `MockOtpFloor` is the mock board's OTP fuse bank — a unary fuse counter, so lowering is impossible by construction, capacity is finite (`Exhausted` past `FUSES`), and `fail_next` injects a one-shot fault. Host integration tests and a QEMU mock-board image link the same crate (`no_std`, cross-builds for `//target/ast10x0`).

Mock only — the `SvnFloor` trait is already on main.

Part of 9elements/openprot#18

### DIFF
--- a/target/mock/BUILD.bazel
+++ b/target/mock/BUILD.bazel
@@ -1,7 +1,7 @@
 # Licensed under the Apache-2.0 license
 # SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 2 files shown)


embediver (approved):
LGTM

PR #435, opened 2026-08-21
docs: Say platform driver, not shell

chrysh:
PR #357 settled on "platform driver" for the effect-carrying layer around the pure state machine. This sweeps the leftover "shell" wording in doc comments and the sm README. Docs only, no code changes.


leongross (approved):
Good catch, lgtm.

PR #437, opened 2026-08-24
orchestrator: Wire CommitSvnFloor through the platform driver

chrysh:
SvnFloor joins BoardCapabilities, one floor per component slot. The SVN to advance to travels in `Verdict::Authenticated`: verification is the only authenticated read of the manifest, so nothing else may tell the floor where to go. The driver caches it per component, clears it on rejection, and fails closed when asked to commit without it. A `None` slot marks a component that tracks its own SVN (iRoT, or a PLDM device committing internally); the eRoT keeps no second floor for it and its commit is a no-op.

Part of 9elements/openprot#18

### DIFF
--- a/services/orchestrator/driver/src/board.rs
+++ b/services/orchestrator/driver/src/board.rs
@@ -7,6 +7,7 @@
 use openprot_orchestrator_sm::{ComponentId, ComponentKind};
 
 pub use orchestrator_capabilities::{BootControl, BootWatch};
+use orchestrator_capabilities::{Svn, SvnFloor};
 
 /// Access to one component's active firmware image, however it is reached —
 /// interposed flash, a PLDM/MCTP transfer, a RAM copy in tests.
@@ -46,8 +47,8 @@ impl<S: ImageSource> ImageSource for &mut S {
     }
 }
 
-/// Judges a component's firmware image; board wiring decides what
-/// "authentic" means.
+/// Judges a component's firmware image; board wiring decides what counts
+/// as authenticated.
 pub trait Verifier {
     /// The error type reported by this verifier.
     type Error: core::error::Error;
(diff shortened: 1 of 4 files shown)


services/orchestrator/driver/src/board.rs
```diff
@@ -99,6 +107,8 @@ pub trait BoardCapabilities {
     type Verifier: Verifier;
     /// Reset actuation for the managed components.
     type BootControl: BootControl;
+    /// The durable anti-rollback floor of the managed components.
```
leongross: why is this durable? what does this imply?

services/orchestrator/driver/src/board.rs
```diff
@@ -128,5 +138,10 @@ pub struct Board<B: BoardCapabilities, const N: usize> {
     /// `boot_controls[i]` actuates `ComponentId(i)`'s reset, same indexing
     /// as `images`.
     pub boot_controls: [B::BootControl; N],
+    /// `svn_floors[i]` is `ComponentId(i)`'s durable anti-rollback floor,
+    /// same indexing as `images`. `None` means the component tracks its
```
leongross: If None means self-managed, we must guarantee a full population at initialization.

services/orchestrator/driver/src/driver.rs
```diff
@@ -45,16 +52,30 @@ pub struct PlatformDriver<B: BoardCapabilities, const N: usize> {
     board: Board<B, N>,
     /// Component whose image is staged (source opened) for verification.
     staged: Option<ComponentId>,
+    /// `verified_svn[i]` is the manifest SVN of `ComponentId(i)`'s last
+    /// authenticated image — the only value a floor commit may trust.
+    /// `None` until a verification passes; cleared again on rejection.
+    verified_svn: [Option<Svn>; N],
```
leongross, line 63: Please correct me if I'm wrong, but both the board capabilites as well as the verified_svn track SVNs for the downstream devices, right? So Why does a `None` in board capabilities resemble a self-managed device, and a `None` here must imply "empty until verification passes"? This is misleading and kind of breaks the API for me. We should find clearer ways to indicate what we want to say to prohibit future misuse.

chrysh: Good catch, that's indeed confusing. Fixed.

services/orchestrator/driver/src/board.rs
chrysh: Durable means it survives power loss and reboot. Fixed in doc comment.

services/orchestrator/driver/src/board.rs
chrysh: Done

leongross (approved):

PR #443, opened 2026-08-26
orchestrator: Compose boot-walk supervision from the board's BootWatch capability

chrysh:
## What

Composes the `BootWatch` capability into the platform driver, closing the gap noted at the `ReleaseReset` executor: released devices are now supervised, and their boot outcomes come back as `ComponentReady`/`Booted`/`Timeout` events.

## How

- `ReleaseReset` arms the component's walk; `AssertReset` stops it, so a device put back in reset (recovery, isolation) can't produce a stale `Timeout`.
- The new `PlatformDriver::poll_boot_walks(now_millis)` polls every watched walk and maps the first terminal verdict to its event: `Complete` becomes `ComponentReady` (Active) or `Booted` (Passive), `Failed` becomes `Timeout` regardless of cause — retry budgeting stays in the SM. A finished walk stops being watched, so each verdict is delivered once. While everything waits, the poll carries the earliest walk deadline as the run loop's next wake-up.
- `BootWatch` gains `arm()`: a retry re-release starts a fresh walk, and since reset actuation has no clock, the attempt starts at the next `poll`'s `now_millis`.
- `Board` gains `boot_watches` and `kinds` (one per component, from the same table as the SM's chain).
- The driver README still described `take_event` and `NotImplemented`, both gone; rewritten to the current execute/fail-closed shape.

## Tests

Seven new driver tests: verdict-to-event mapping by kind, exactly-once delivery, earliest-deadline aggregation, watch gating on release/assert, fresh-walk re-arming, and two end-to-end runs through the SM (passive Booted settles in Ready; boot timeout fails closed while the recovery seam is uncomposed).

## Known gap

The walk can report `FailureCause::DeviceFatal`, but the SM's event vocabulary only has `Timeout`, so fatal causes still spend retry budget. Short-circuiting those needs a new SM event — follow-up, not this PR.

### DIFF
--- a/services/orchestrator/driver/src/driver.rs
+++ b/services/orchestrator/driver/src/driver.rs
@@ -4,10 +4,10 @@
 //! The [`PlatformDriver`]: one executor method per [`Effect`] variant, routed from
 //! the SM through the [`Platform`] impl.
 
-use openprot_orchestrator_sm::{ComponentId, Effect, EffectError, Event, Platform};
+use openprot_orchestrator_sm::{ComponentId, ComponentKind, Effect, EffectError, Event, Platform};
 
 use crate::board::{Board, BoardCapabilities, ImageSource, Verdict, Verifier};
-use orchestrator_capabilities::BootControl;
+use orchestrator_capabilities::{BootControl, BootWatch, WalkVerdict};
 
 /// Why the driver could not carry out an effect.
 #[derive(Clone, Copy, PartialEq, Eq, Debug)]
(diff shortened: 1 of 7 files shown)


services/orchestrator/driver/src/driver.rs
```diff
@@ -101,22 +107,95 @@ impl<B: BoardCapabilities, const N: usize> PlatformDriver<B, N> {
...
+    /// [`BootWalkPoll::event`] is `None`. Only that last poll carries a
+    /// complete [`next_deadline_millis`](BootWalkPoll::next_deadline_millis)
+    /// — the earliest deadline among the still-waiting walks.
+    pub fn poll_boot_walks(&mut self, now_millis: u64) -> BootWalkPoll {
+        let mut next_deadline_millis: Option<u64> = None;
+        for idx in 0..N {
+            if !self.watching[idx] {
+                continue;
+            }
+            let id = ComponentId::new(idx as u8);
```
leongross: nit: a bit bulky; consider implementing `From<u8>` for easy into casts:

```rust
impl From<u8> for ComponentId {
    fn from(value: u8) -> Self {
        Self(value)
    }
}

impl From<ComponentId> for u8 {
    fn from(value: ComponentId) -> Self { value.0 }
}
```

leongross (approved):

leongross:
I would feel better if we could get rid of the index matchings, but as @chrysh already discussed, we will leave it like this until @FerralCoder finishes the config design.

PR #444, opened 2026-08-26
orchestrator: Fold post-write readback into the Updatable contract

chrysh:
`Ready` now means the staged payload is verified to the archetype's discipline: a direct-flash adapter reads written pages back before reporting `Ready`, a PLDM device verifies inside its own transfer (PLDM Type 5 has no command to read a slot back; its Verify step runs before apply). This drops the planned `ReadBack` capability: a post-`Ready` check would leak archetype knowledge to the caller.

`MockFlashDevice` demonstrates the discipline: write steps hold `written` still, readback advances it, so `written` counts verified bytes; a mismatch is a staging error.

References 9elements/openprot#8.

### DIFF
--- a/services/orchestrator/capabilities/src/updatable.rs
+++ b/services/orchestrator/capabilities/src/updatable.rs
@@ -24,12 +24,13 @@
 /// What this trait deliberately does not claim:
 ///
 /// - **Verification** runs on the candidate before staging as
-///   orchestrator policy; post-write read-back is the optional
-///   `ReadBack` capability.
+///   orchestrator policy. Post-write readback is no capability either:
+///   it is the implementor's staging discipline (see `Ready` below).
 /// - **Commit.** Activation is always tentative: it proposes the staged
 ///   image as the preferred boot target, never commits it. The commit
-///   gate is the optional `TrialBoot` capability, which resolves what
-///   `activate` proposed; there is no second slot-selection owner.
+///   gate is the orchestrator's confirmed-boot flow (`BootConfirmed`
+///   gating `SvnFloor::advance`, or the device committing internally);
+///   there is no second slot-selection owner.
 /// - **Booting.** Resetting the device into the candidate is
 ///   [`BootControl`](crate::BootControl). When activation takes effect
 ///   (next reset, or a device-internal restart on self-activating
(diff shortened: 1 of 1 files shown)


services/orchestrator/capabilities/src/updatable.rs
```diff
@@ -597,28 +615,37 @@ mod tests {
...
+    // `Ready` contract prescribes for direct-flash devices.
     const FLASH_PAGE: usize = 2; // bytes programmed per write step
     const FLASH_SECTOR: usize = 4; // bytes erased per erase step
     const FLASH_SECTORS: usize = 2; // sectors in the slot
 
     struct MockFlashDevice {
         slot: [u8; FLASH_SECTORS * FLASH_SECTOR],
         erased: [bool; FLASH_SECTORS],
+        // The page written last step, awaiting readback: (start, expected).
+        pending: Option<(usize, Vec<u8>)>,
```
leongross: Personally, I am not a big fan of explicit inline tuples for such data representation.

Maybe something like 
```rust
pub struct ExpectedReadback(u8, Vec<u8>);
```

works for this? Feel free to choose a better-fitting name here.

leongross (approved):

services/orchestrator/capabilities/src/updatable.rs
```diff
@@ -52,9 +53,15 @@
 ///   [`abandon`](Self::abandon) instead of waiting out a blocked call.
 ///   Liveness policy stays with the caller: it watches `written` and
 ///   abandons a transfer that stalls too long, on its own clock.
-/// - **`Ready` means ready.** The device holds the complete payload and
-///   `activate` may be called. `activate` in any other staging state is
-///   an error.
+/// - **`Ready` means ready.** The device holds the complete payload,
+///   verified to its archetype's discipline: a direct-flash adapter
+///   reads written data back and re-verifies before reporting `Ready`,
+///   a PLDM device runs its own verify step inside the transfer. A
```
rusty1968: Thhe comment claims a PLDM device verifies internally, but nothing in this PR (or the existing PLDM mock) actually shows that. It would be nice to model this in a test.

rusty1968 (approved):

services/orchestrator/capabilities/src/updatable.rs
rusty1968: Yep this is harder to read.

services/orchestrator/capabilities/src/updatable.rs
```diff
@@ -751,6 +783,20 @@ mod tests {
...
+        dev.corrupt_next_write = true;
+        let payload = SliceSource(b"8 bytes!");
+
+        let err = stage_all(&mut dev, &payload).expect_err("expected the readback mismatch");
+        assert_eq!(err, UpdateError::ReadbackMismatch);
+
+        // Staging anew after the fault is allowed and verifies clean.
+        stage_all(&mut dev, &payload).expect("re-staging failed");
+        dev.activate().expect("activate after re-staging failed");
+    }
```
rusty1968, line 830: never reads dev.slot back to confirm it actually equals b"8 bytes

rusty1968:
@CourtneyDrant  can you review as well?

services/orchestrator/capabilities/src/updatable.rs
chrysh: That's true!

CourtneyDrant (approved):

PR #445, opened 2026-08-26
Orchestrator: gracefully gate components the platform can't recover

rusty1968:
The intent of this PR is to allow the platform to boot degraded when a non-required (Isolable/Cascading) component cannot be recovered, instead of hard locking.

### DIFF
--- a/docs/src/design/orchestrator/orchestrator-model.md
+++ b/docs/src/design/orchestrator/orchestrator-model.md
@@ -65,11 +65,16 @@ verification.
 
 ```
 Required   — attempt recovery; re-walk the chain after `Restored`; latch
-             `Locked` if the retry cap is reached
+             `Locked` once recovery is exhausted
 Isolable   — hold the component in reset; advance past it; continue the walk
 Cascading  — same as Isolable, and additionally hold in reset any component
              whose `depends_on` names this one
 ```
+
+Recovery is *exhausted* either when the retry cap is reached or when the
+platform reports `RecoveryUnavailable` — both run the same policy branch, so a
+platform that is out of recovery images gates an `Isolable` component rather
+than halting the platform.
 
 ### `RegionId`
 
@@ -361,7 +366,10 @@ names **what** must happen to **which** component; the platform decides **how**.
 For example, `Effect::RecoverComponent(id)` says only "recover this component" —
 whether that resolves to a golden-image restore, an A/B slot swap, a streamed
 image, or a vendor-specific scheme is a platform/configuration decision, never
-encoded in the core.
+encoded in the core. The platform reports the outcome as an event: `Restored`
+once it has swapped in an untried image, `RecoveryUnavailable` when it has none
+left. Both are verdicts; `EffectError` stays reserved for a genuine actuation
+fault, which fails closed to `Locked` regardless of policy.
 
 The core never reads flash, never checks signatures, never observes reset lines.
 It only emits descriptions. The complete split:
--- a/services/orchestrator/sm/src/lib.rs
+++ b/services/orchestrator/sm/src/lib.rs
@@ -424,6 +424,20 @@ impl<const N: usize, const E: usize> Rot<N, E> {
         }
     }
 
+    /// Whether a `VerificationPassed(id)` may release `id`. Two conditions, and
+    /// both release sites check them through here so they cannot drift apart:
+    ///
+    /// - `id` is the component currently under verification (`chain[cursor]`,
+    ///   whose `VerifyFirmware` was just emitted). A verdict for any other id is
+    ///   stale or out-of-turn (cf. INV9 for `ComponentReady`).
+    /// - `id` is not gated. Gating does not move the cursor off the component,
+    ///   so a component isolated *while its verification was in flight* is still
+    ///   `chain[cursor]` when that verdict lands — and a stale pass must not
+    ///   undo the `AssertReset` that just held it (INV8).
+    fn may_release(&self, id: ComponentId) -> bool {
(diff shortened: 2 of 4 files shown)


chrysh:
`may_release` fixes the INV8 hole, but it also removes the only thing that kept the walk alive when `chain[cursor]` is gated mid-verification. `handle_corruption`'s `Gated` arm returns `Handled` with the cursor untouched, so once the stale pass is dropped, no event can advance the walk anymore: `PreSupervision` has no supervisor, so `Timeout` is discarded, and in `AwaitingReady` the later `ComponentReady` sees `cursor < chain.len()` and settles in `AwaitingReady(None)`. The platform never reaches `Ready`. That's a boot halted on a component the board classified `Isolable` — exactly what this PR is supposed to prevent.

The new `stale_verdict_*` tests don't catch it because they bind `_state`. These two fail on the current branch, the other 87 pass:

```rust
/// A component gated while its verification is in flight must not park the
/// walk: the rest of the chain still verifies and the platform boots degraded.
#[test]
fn walk_completes_past_component_gated_under_verification() {
    let (effects, state) = drive(
        chain(&[
            (C0, ComponentAttrs::passive_required()),
            (C1, ComponentAttrs::passive_isolable()),
            (C2, ComponentAttrs::passive_required()),
        ]),
        &[
            BOOT,
            Event::VerificationPassed(C0), // → VerifyFirmware(C1) in flight
            Event::CorruptionDetected(C1), // → AssertReset(C1), isolated
            Event::VerificationPassed(C1), // the in-flight verdict, now stale
            Event::VerificationPassed(C2),
        ],
    );
    assert!(effects.contains(&Effect::VerifyFirmware(C2))); // fails: never emitted
    assert_eq!(state, State::Ready); // fails: parked in PreSupervision
}

/// Same hole in `AwaitingReady`: `ComponentReady` must finish the walk, not
/// settle in `AwaitingReady(None)` forever.
#[test]
fn awaiting_ready_completes_past_component_gated_under_verification() {
    let (effects, state) = drive(
        chain(&[
            (C0, ComponentAttrs::active_required()),
            (C1, ComponentAttrs::passive_isolable()),
        ]),
        &[
            BOOT,
            Event::VerificationPassed(C0), // → AwaitingReady(C0); VerifyFirmware(C1) in flight
            Event::CorruptionDetected(C1), // → AssertReset(C1), isolated
            Event::VerificationPassed(C1), // the in-flight verdict, now stale
            Event::ComponentReady(C0),
        ],
    );
    assert!(!effects.contains(&Effect::ReleaseReset(C1)));
    assert_eq!(state, State::Ready); // fails: stuck in AwaitingReady(None)
}
```

The verdict for a gated component might never arrive at all, so the re-arm belongs at the gating site, not the drop site: advance the walk when gating lands on `chain[cursor]`. And `AwaitingReady`'s done-check needs to be "no ungated component at or after cursor" instead of `cursor < chain.len()`.

chrysh:
orchestrator-machine.md seems to need updating (e.g. RecoveryUnavailable)

chrysh (changes requested):
`may_release` makes the stale pass inert, but the stale fail for the same gated component is still honored: `VerificationFailed(C1)` right after `CorruptionDetected(C1)` enters `Recovering(C1)` and emits `RecoverComponent` for a component policy just took out of service. Entering recovery also runs `quiesce_all`, so every healthy live component is reset for a device the re-walk will skip anyway. The end state self-heals — the re-walk skips the gated component — but we rewrite the flash of an isolated device and re-boot the whole platform for nothing. And with a driver that answers `RecoverComponent` with `Err`, this stale fail escalates all the way to `Locked`.

Shouldn't a verdict for a gated component be inert in both polarities, not just the pass? Same hole in the `AwaitingReady` arm.

This fails on the current branch:

```rust
/// A verdict for a gated component is inert regardless of polarity: the stale
/// `VerificationFailed` must not start a recovery episode for a component
/// policy just took out of service.
#[test]
fn stale_verdict_never_recovers_an_isolated_component() {
    let (effects, state) = drive(
        chain(&[
            (C0, ComponentAttrs::passive_required()),
            (C1, ComponentAttrs::passive_isolable()),
        ]),
        &[
            BOOT,
            Event::VerificationPassed(C0), // → VerifyFirmware(C1) in flight
            Event::CorruptionDetected(C1), // → AssertReset(C1), isolated
            Event::VerificationFailed(C1), // the in-flight verdict, now stale
        ],
    );
    assert_ne!(state, State::Recovering(C1)); // fails: episode started
    assert!(!effects
        .iter()
        .any(|e| matches!(e, Effect::RecoverComponent { id, .. } if *id == C1)));
    // Collateral: entering `Recovering` quiesces the already-released C0.
    assert!(!effects.contains(&Effect::AssertReset(C0)));
}
```

rusty1968:
main picked up b18b78c ("Move the cursor off a component gated mid-walk") and 5060c36 ("Drop verdicts for a component already isolated") after this branch forked — same fixes this review was asking for, but taking the fix at the gating site instead of the release-check site, which avoids the walk-hang may_release introduced here. Confirmed directly: both repro tests from the first comment (walk_completes_past_component_gated_under_verification, awaiting_ready_completes_past_component_gated_under_verification) and the repro test from the CHANGES_REQUESTED review (stale_verdict_never_recovers_an_isolated_component) all pass unmodified against current main.

The actual net-new contribution here — RecoveryUnavailable + exhaust_recovery — carries forward cleanly rebased on top of those fixes, and the follow-up also covers the two things this PR was still missing: RecoveryUnavailable wired into the random_event property-test fuzzer, and the orchestrator-machine.md update from the second comment.

Superseded by #473

PR #447, opened 2026-08-27
orchestrator: Add the ReportSink seam for the Report effects

chrysh:
The four `Effect::Report*` variants have no consumer seam. This adds a `#[non_exhaustive]` `Report` enum and a single `ReportSink::report` instead of a method or a trait per variant, so a new report is one variant rather than a trait break and a sink routes only what it recognises.

`report` returns nothing: a report names something that already happened, so an undeliverable one costs information, not containment, and reporting stays off the driver's fail-closed path. Payloads stay `Copy` and lifetime-free, like the effects they mirror.

The seam lives in `driver/src/board.rs` rather than `orchestrator-capabilities`, because `Report` carries a `ComponentId` and that crate is a dependency-free leaf. Same split the driver already makes for `Verifier`.

Composing a sink into the board follows in the next PR.

References 9elements/openprot#3.

### DIFF
--- a/services/orchestrator/driver/src/board.rs
+++ b/services/orchestrator/driver/src/board.rs
@@ -89,6 +89,59 @@ pub enum Verdict {
     Rejected,
 }
 
+/// One fact about the platform running degraded, carried outward to
+/// management software. `#[non_exhaustive]`: a sink routes what it
+/// recognises and ignores the rest, so a new report is not a trait break.
+///
+/// Payloads stay `Copy` and lifetime-free, like the effects these mirror. A
+/// report names a component only where the effect it mirrors does.
+#[non_exhaustive]
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum Report {
+    /// Held in reset and out of the trust chain, so the platform runs
+    /// degraded. Reported once per component, as it is gated.
+    Isolated(ComponentId),
+    /// Recovery attempts exhausted and the platform halts. Reported before
+    /// the lockdown latch, while there is still a platform to report from.
+    RecoveryFailed(ComponentId),
+    /// An update request declined because the platform was busy. Nothing
+    /// staged, and the requester may ask again. Platform-wide, not
+    /// per-component: the machine supervises one update at a time and
+    /// `Event::UpdateRequest` names no component.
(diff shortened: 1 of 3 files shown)


services/orchestrator/driver/src/board.rs
```diff
@@ -89,6 +89,84 @@ pub enum Verdict {
...
+    /// losing its reports. Implementations saturate at `u32::MAX` rather
+    /// than wrapping. A variant the sink does not route is not a loss: it
+    /// never had a destination, the same reason `()` counts none.
+    ///
+    /// No default: a sink that can lose a report has to say how many, and one
+    /// that cannot returns 0 explicitly.
+    ///
+    /// [`report`]: ReportSink::report
+    fn dropped(&self) -> u32;
+}
```
leongross: I do not fully understand how this information is beneficial to us. Can you please explain it briefly?

rusty1968: This seems to be an a hook  for platform management software.

services/orchestrator/driver/src/board.rs
```diff
@@ -89,6 +89,84 @@ pub enum Verdict {
...
+    /// management side has no other way to tell a quiet platform from one
+    /// losing its reports. Implementations saturate at `u32::MAX` rather
+    /// than wrapping. A variant the sink does not route is not a loss: it
+    /// never had a destination, the same reason `()` counts none.
+    ///
+    /// No default: a sink that can lose a report has to say how many, and one
+    /// that cannot returns 0 explicitly.
+    ///
+    /// [`report`]: ReportSink::report
+    fn dropped(&self) -> u32;
```
rusty1968: Cut dropped() — no sink needs loss-counting yet; add it when one does.

chrysh: Ok

services/orchestrator/driver/src/board.rs
chrysh: Do we want a reported counter? I feel like having some counter could be useful to see whether anything unusual is happening on the platform?

@rusty1968 If not, can you approve and merge this PR?

rusty1968 (approved):

PR #448, opened 2026-08-27
orchestrator: Compose report delivery from the board's ReportSink capability

chrysh:
TL;DR Wire reportsink to board

`BoardCapabilities` gains a `ReportSink` and `Board` a `report_sink` field.
One sink per platform, not one per component: two of the four reports name
no component. The four `Report` effects leave the not-yet-composed group and
hand their `Report` to the sink.

Reporting returns no error, so it never reaches the fail-closed group. An
`Isolable` component is contained, reported, and the platform stays `Ready`.

References 9elements/openprot#3.


rusty1968:
Move from draft to ready for review.

chrysh:
Moved back to draft until 447 is merged

leongross (approved):
Looks good to me!

rusty1968 (approved):

PR #449, opened 2026-08-27
docs: add orchestrator runtime concept page and board composition

rusty1968:
Introduces a design page on the orchestrator runtime and how it turns hardware and IPC activity into the events the state machine consumes.

### DIFF
--- a/docs/src/design/orchestrator/orchestrator-runtime.md
+++ b/docs/src/design/orchestrator/orchestrator-runtime.md
@@ -1,0 +1,111 @@
+# Runtime
+
+The **runtime** is the orchestrator's event loop: the single task that sits
+between two collaborators — the pure [state machine](./orchestrator-machine.md)
+(the *core*) that decides, and the `PlatformDriver` that executes — turning
+outside happenings (interrupts, timeouts, IPC messages) into the `Event`s the
+core consumes, and handing the `Effect`s it emits to the driver. It holds no
+policy of its own — every decision stays in the core; the runtime only moves
+information across the process boundary.
+
+It gathers three possible inbound sources — **hardware interrupts**
+(boot-progress lines the orchestrator owns directly), **IPC channels**, and
+**watchdog deadlines** (timeouts) — into the one event stream the core
+supervises, and carries the core's decision back out. Which sources are present
+is a [board-composition](./orchestrator-overview.md#board-composition) choice:
+boot-progress is a hardware interrupt only when the orchestrator owns the pins;
+otherwise a monitor forwards it as an IPC message.
+
+The unifying idea is the kernel **wait group**: interrupts and IPC are not
+separate mechanisms but interchangeable *members* of one group, and a watchdog
+deadline is that same wait's *timeout* — so all three collapse into the return
+of one `object_wait(handle, signal_mask, deadline)`. A *signal* here is just a
+named bit on a waitable object, not a source in its own right: a latched IRQ bit
+on an interrupt object, or `READABLE` / `USER` on a channel (a service raises
+`Signals::USER` on a client channel to notify without a reply). The loop is
+written once against "a member that signaled, or the deadline that lapsed,"
+never against a specific source; what differs between sources is only the
+*decoder* that turns each into an `Event`.
+
+Two neighboring pages carry the supporting detail:
+[Platform Architecture](./orchestrator-platform.md) names the services and the
+responsibility boundary, and [pw_kernel IPC](../pw-kernel-ipc.md) gives the
+concrete channel syscalls. The worked, compiling reference is the QEMU
+integration test at `target/ast10x0/tests/orchestrator/runtime/main.rs`.
+
+## The single wait point
+
+The runtime is a single-threaded loop parked in one place: a kernel
+`object_wait` over a **wait group**. Every inbound source is registered once as a
+*member* of that group (`wait_group_add`), and the wait returns whichever member
+signaled. Each member resolves to at most one `Event`:
+
+- **Boot-progress signals** — a component reaching a checkpoint raises a
+  boot-progress signal. Depending on the board's composition it arrives as an
+  interrupt object the orchestrator holds directly or as a message a monitor
+  forwards over a channel; either way the loop maps it to
+  [`Event::ComponentReady`] (an `Active` component's iRoT-verified readiness)
+  or [`Event::Booted`] (a `Passive` component's liveness).
+- **Watchdog deadlines** — the timer is not a separate task. `BootWatchdogs`
+  (`services/orchestrator/server`) folds all armed boot windows and the commit
+  window into a *single deadline* passed straight to `object_wait`. When the
+  wait returns `DeadlineExceeded`, `poll_expired()` yields the mapped
+  [`Event::Timeout`] / [`Event::CommitTimeout`].
+- **IPC channel messages** — an update agent, management path, peer service, or
+  a boot-progress-forwarding monitor holds a channel *initiator*; the runtime
+  holds the *handler*. A readable channel is another object the loop waits on;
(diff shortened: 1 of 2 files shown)


chrysh:
Add to docs/src/SUMMARY.md:
```
* [Runtime](./design/orchestrator/orchestrator-runtime.md)     
```

chrysh:
The runtime page says "The three share one `object_wait`, so a slow image hash on one path cannot delay a boot window on another". I don't think the wait group gives us that.

`object_wait` bounds how long the loop *waits*, not how long it *works*. While an effect executor runs, the loop is not in the wait at all, so no deadline is in effect. Verify is inline today: `Effect::VerifyFirmware` calls `Verifier::verify` in `driver/src/driver.rs:86`. With a 200ms boot window and a 2s image hash, the window lapses at 200ms, nobody notices until the hash returns at 2s, and `Event::Timeout` fires 1.8s late.

What actually keeps the loop live is the other half, which the page does not state: every decode and every effect executor returns promptly. That is why staging is a pump rather than a single transfer call. Could the page say that rule directly instead of deriving liveness from the shared wait?

Two related questions:

- The platform page says commands to crypto and storage are fire-and-forget and every reply comes back as a queued event. `channel_transact` is synchronous and it is the only outbound primitive the kernel gives us. What is the intended shape there?
- Sources with no signal (the `BootWatch` walkers, later the update pump) need the loop to wake on its own. Would an adaptive deadline cover it: `now` while a job is live, `wait_deadline()` when idle?

chrysh:
The page says `BootWatchdogs` folds all armed boot windows and the commit window into a single deadline. There is a second tracker for the same windows though: `release_reset` arms a `BootWatch` per component, and `poll_boot_walks(now_millis)` returns `next_deadline_millis` plus `Event::Timeout(id)` on `WalkVerdict::Failed` (`driver/src/driver.rs:150`).

So today the same boot window is tracked twice, in two units (`Instant` in `BootWatchdogs`, `u64` millis in the walkers), and both paths produce `Event::Timeout(id)`. The state machine drops the second one as stale via `is_awaiting_boot`, so nothing breaks, but two owners of one deadline is the failure mode we already agreed to avoid elsewhere.

The part that does break: the walkers are polled and nothing signals the loop on their behalf. If `wait_deadline()` only reflects `BootWatchdogs`, the loop sleeps straight past a walker's deadline.

Which one is meant to survive?

chrysh:
poll_boot_walks(now_millis) returns Event::Timeout(id) when a walk hits WalkVerdict::Failed (services/orchestrator/driver/src/driver.rs:150), so there is an edge from the driver back into dispatch that the picture does not show. Worth an arrow?

docs/src/design/orchestrator/orchestrator-runtime.md
```diff
@@ -0,0 +1,111 @@
...
+The **runtime** is the orchestrator's event loop: the single task that sits
+between two collaborators — the pure [state machine](./orchestrator-machine.md)
+(the *core*) that decides, and the `PlatformDriver` that executes — turning
+outside happenings (interrupts, timeouts, IPC messages) into the `Event`s the
+core consumes, and handing the `Effect`s it emits to the driver. It holds no
+policy of its own — every decision stays in the core; the runtime only moves
+information across the process boundary.
+
+It gathers three possible inbound sources — **hardware interrupts**
+(boot-progress lines the orchestrator owns directly), **IPC channels**, and
```
leongross, line 12: Is the boot monitoring really steered by hardware interrupts? Our trait based implementation as of now is naturally agnostic, but when we discussed this we talked about polling the GPIO lines, instead of waiting for real hardware interrupts handled by the hardware NVIC and ISRs.

I understand what you want to say by this, but if we are not relying on hardware inerrupts we should adapt the wording here.

docs/src/design/orchestrator/orchestrator-runtime.md
```diff
@@ -0,0 +1,111 @@
...
+The unifying idea is the kernel **wait group**: interrupts and IPC are not
+separate mechanisms but interchangeable *members* of one group, and a watchdog
+deadline is that same wait's *timeout* — so all three collapse into the return
+of one `object_wait(handle, signal_mask, deadline)`. A *signal* here is just a
+named bit on a waitable object, not a source in its own right: a latched IRQ bit
+on an interrupt object, or `READABLE` / `USER` on a channel (a service raises
+`Signals::USER` on a client channel to notify without a reply). The loop is
+written once against "a member that signaled, or the deadline that lapsed,"
+never against a specific source; what differs between sources is only the
+*decoder* that turns each into an `Event`.
```
leongross, line 28: What does this decoder look like and where is it defined?

docs/src/design/orchestrator/orchestrator-runtime.md
```diff
@@ -0,0 +1,111 @@
...
+responsibility boundary, and [pw_kernel IPC](../pw-kernel-ipc.md) gives the
+concrete channel syscalls. The worked, compiling reference is the QEMU
+integration test at `target/ast10x0/tests/orchestrator/runtime/main.rs`.
+
+## The single wait point
+
+The runtime is a single-threaded loop parked in one place: a kernel
+`object_wait` over a **wait group**. Every inbound source is registered once as a
+*member* of that group (`wait_group_add`), and the wait returns whichever member
+signaled. Each member resolves to at most one `Event`:
```
leongross, line 41: So what is the use case for a registered member to resolve to no Event?
If this is not the case, we should be precise and say 
```suggestion
signaled. Each member maps to exactly one `Event`:
```

docs/src/design/orchestrator/orchestrator-runtime.md
```diff
@@ -0,0 +1,111 @@
...
+**watchdog deadlines** (timeouts) — into the one event stream the core
+supervises, and carries the core's decision back out. Which sources are present
+is a [board-composition](./orchestrator-overview.md#board-composition) choice:
+boot-progress is a hardware interrupt only when the orchestrator owns the pins;
+otherwise a monitor forwards it as an IPC message.
+
+The unifying idea is the kernel **wait group**: interrupts and IPC are not
+separate mechanisms but interchangeable *members* of one group, and a watchdog
+deadline is that same wait's *timeout* — so all three collapse into the return
+of one `object_wait(handle, signal_mask, deadline)`. A *signal* here is just a
```
leongross, line 22: Could we link the source defintiion/link to pigweed kernel upstream sources here?

rusty1968:
> The runtime page says "The three share one `object_wait`, so a slow image hash on one path cannot delay a boot window on another". I don't think the wait group gives us that.
> 
> `object_wait` bounds how long the loop _waits_, not how long it _works_. While an effect executor runs, the loop is not in the wait at all, so no deadline is in effect. Verify is inline today: `Effect::VerifyFirmware` calls `Verifier::verify` in `driver/src/driver.rs:86`. With a 200ms boot window and a 2s image hash, the window lapses at 200ms, nobody notices until the hash returns at 2s, and `Event::Timeout` fires 1.8s late.
> 
> What actually keeps the loop live is the other half, which the page does not state: every decode and every effect executor returns promptly. That is why staging is a pump rather than a single transfer call. Could the page say that rule directly instead of deriving liveness from the shared wait?
> 
> Two related questions:
> 
> * The platform page says commands to crypto and storage are fire-and-forget and every reply comes back as a queued event. `channel_transact` is synchronous and it is the only outbound primitive the kernel gives us. What is the intended shape there?
> * Sources with no signal (the `BootWatch` walkers, later the update pump) need the loop to wake on its own. Would an adaptive deadline cover it: `now` while a job is live, `wait_deadline()` when idle?

This is what we discussed this morning - we don't need to wait for long running effects to be completed inline. We can dispatch them and expect completion events to be fed back as notifications (handler signals and we fetch the results).

PR #455, opened 2026-09-01
orchestrator: Add the update intake seam for an update source

chrysh:
The IPC channel between `services/pldm` and the orchestrator, as its contract.

`wire.rs` is the channel protocol: a 16-byte request header plus the chunk on
`Write`, a 20-byte response, five opcodes, and `MAX_CHUNK = 512` to match
`FD_MAX_XFER_SIZE` in `pldm-interface` so a `RequestFirmwareData` chunk passes
through unsplit. Decoding validates lengths, opcodes, reserved fields and the
chunk bound before anything looks at the payload, and never panics: the
request side is another process and is treated as untrusted.

`traits.rs` is the `UpdateIntake` call surface, `status.rs` the `IntakeStatus`
phases the orchestrator answers with. Host-buildable, depends only on
zerocopy, and both processes depend on it.

The source is the channel's initiator and the orchestrator its handler, with
no channel the other way, so the orchestrator never waits on a source and a
wedged source cannot delay a boot window or a recovery.

Not here, and why:

- The `channel_transact` client and the `channel_read`/`channel_respond`
  handler loop are separate PRs, trait first.
- The `system.json5` endpoint declarations cannot land yet:
  `target/ast10x0/erot/system.json5` is kernel-only and has no `apps:` key, so
  there are no processes to hang a `channel_handler` and `channel_initiator`
  off.
- `services/storage` is still a stub, so nothing moves a real image. The crate
  and its tests stand alone on the host.

Draft because one thing is unresolved: `Event::UpdateRequest` fires at
`RequestUpdate` accept on main (#442), which enters `State::Updating` and
emits `AuthenticateUpdate` with no bytes to authenticate. The README no longer
claims `complete()` is what hands the state machine the request, but where
that event should fire is not settled here.

Stacked on Anthony's runtime-concept commit (51b4dfa), which is not merged.
Review the last two commits only.

### DIFF
--- a/docs/src/design/orchestrator/orchestrator-overview.md
+++ b/docs/src/design/orchestrator/orchestrator-overview.md
@@ -1,4 +1,4 @@
-# Orchestrator State Machine
+# Orchestrator
 
 The orchestrator is the eRoT's boot-sequence controller. It walks the platform
 trust chain — verifying each component's firmware and releasing it from reset in
@@ -28,6 +28,10 @@ avoid the two drifting apart.
 - [**Platform Architecture**](./orchestrator-platform.md): The platform half around
   the core — surrounding services, capability contracts, the board device table,
   and the fail-safe rules at the responsibility boundary.
+- [**Runtime**](./orchestrator-runtime.md): How the runtime loop
+  gathers hardware interrupts, IPC channel messages, and watchdog deadlines into
+  the core's event stream, and carries the resulting effects and decisions back
+  out.
 
 ## Design Principles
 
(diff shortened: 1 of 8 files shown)


chrysh:
Closing this. The intake seam, handler, and loopback harness from this PR now live on `pldm-update-request-ipc`, rebased onto current main together with the pldm-notify wire, the notify handler, and the Updatable readback fold. The stacked PRs from that branch will carry the work in.

The open question in the description (where `Event::UpdateRequest` fires) is also addressed there: the notify channel replaced the `complete()`-fires-it path.

chrysh:
Reopening. The close comment above was wrong: `pldm-update-request-ipc` carries only the pldm-notify channel, not the intake seam. The wire protocol, `UpdateIntake` trait, handler and loopback harness in this PR exist nowhere else. Where `Event::UpdateRequest` fires is still the open question the description states.

chrysh:
Closing this. #464 settled the direction the other way around: PLDM-FD is the
IPC server and the orchestrator is the client that grants or denies. An intake
seam where the source pushes offer/write/complete into the orchestrator does
not match that, and most of this PR is the transfer path on the wrong side.

What carries over goes into the orchestrator-to-FD API when that gets built:
the deny reasons (unknown target, busy, out of range, plus deferred and
superseded as failure causes), the split between a peer that answered and
refused and a transport that gave no answer, and reserving staging at accept
time.

The runtime docs page in the first commit is Anthony's, it is #449.

PR #456, opened 2026-09-02
Initial draft of OCP Global Demo Proposal.

CourtneyDrant:


chrysh:
pr456-demo-flow
Looking at the diagram, how can the BMC be informed about an invalid image when it is not booted yet?

chrysh (changes requested):
See my previous comment

chrysh:
How does the BMC signal to the eRoT a successfull boot? How do we distinguish it from a hung BMC?

What happens on failure?

@FerralCoder

chrysh:
Maybe do 12 (Mark B as active) first, then 11 (Erase staging area). Otherwise, in case there is a power outage between 11 and 12, we would need to retransfer the firmware over PLDM.
If we do 12, then 11: Compare the staged digest against active B digest and erase staging area if disgests are equal, or keep a separate staging applied to slot B marker.

Or maybe even better: Erase staging area after successful boot?

chrysh:
Small issues:
- OpenPRoT: Fix spelling
- AST2700 or AST2070?
- Is ActivatePendingComponentImage a PLDM T5 command? I assume the black lines are PLDM commands?
- Is there a timeout between tell BMC to shut down and pulling the power? @FerralCoder 

CAD:  Will fix.  ActivatePendingComponentImage is a T5 Command used while FSM is "IDLE" for FW activation without Component Image Transfer.

chrysh:
Why do we do this "Update inactive slot A to match slot B"? Don't we want to keep a "previously known-good firmware" for rollback? What if the BMC turns out to have a broken daemon or service or so?

("requires mastership and ANOTHER BMC power cycle")

@FerralCoder

chrysh:
It looks like in case of image bad, we still end up in good update completed after the BMC boot completes.

chrysh:
Maybe the diagram should start by the BMC writing the new firmware image into the SPI?

chrysh:
Should we also add Timeouts into this sequence diagram, e.g. boot timeout?

docs/demo-proposal-flow.md
```diff
@@ -0,0 +1,47 @@
...
+Use OpenProt as Firmware Device to execute extern staging flow listed below.  
+Update agent will send GetFirmParameters for Inventory and ActivatePendingComponentImage to signal to OpenProt to begin this flow.
+
+sequenceDiagram
+    autonumber
+    participant BMC as BMC
+    participant OpenProt as OpenProt
+    participant fwspi as BMC SPI bus (fwspi)
+
+    BMC->>OpenProt: GetFirmwareParameters
```
rusty1968, line 12: Add the trigger event to the flow - The flow diagram never shows the event triggering the flow - an external agent writing the flash image to the staging area - The first documented event is GetFirmwareParameters, and by the time OpenPRoT does "Verify BMC image in staging area," the diagram just assumes the image is already sitting there.

rusty1968:
@CourtneyDrant , I have proposed some changes to the flow as a PR to your fork.

docs/demo-proposal-flow.md
```diff
@@ -0,0 +1,47 @@
+Propsal:
+Build a small Update Agent leverage pldm-lib/pldm-common to run ontop of OpenBMC on AST2700.
+Use OpenProt as Firmware Device to execute extern staging flow listed below.  
+Update agent will send GetFirmParameters for Inventory and ActivatePendingComponentImage to signal to OpenProt to begin this flow.
+
+sequenceDiagram
```
rusty1968, line 6: There should be more actors here - for the demo there should be a human trigger - the demonstrator opens a webui with a file picker and an upload firmware button.

docs/demo-proposal-flow.md
```diff
@@ -0,0 +1,47 @@
...
+    OpenProt->>fwspi: Claim mastership
+    OpenProt->>OpenProt: Verify BMC image in staging area
+
+    alt Image good
+        OpenProt->>OpenProt: Copy staging image to "B" partition
+        OpenProt->>OpenProt: Verify good copy of "B"
+        OpenProt->>OpenProt: Erase staging area
+        OpenProt->>OpenProt: Mark "B" partition as active
+    else Image bad
+        OpenProt-->>BMC: Activation failed (image invalid)
```
rusty1968, line 34: OpenProt tries to send "Activation failed (image invalid)" to a BMC that has no power.

docs/demo-proposal-flow.md
```diff
@@ -0,0 +1,47 @@
...
+
+    opt Optional query
+        BMC->>OpenProt: QueryDeviceIdentifiers
+        OpenProt-->>BMC: Descriptors
+    end
+
+    BMC->>OpenProt: ActivatePendingComponentImage(AST2070 Component Identifier)
+    OpenProt-->>BMC: EstimatedTimeForActivation
+
+    Note over OpenProt,BMC: OpenProt disables access by BMC (notify BMC to shutdown, then pull power)
```
rusty1968, line 23: What is the transport for this notification? Currently openBMC and the AST1060 communicate via a software-based mailbox over the I2C. What is the

docs/demo-proposal-flow.md
```diff
@@ -0,0 +1,47 @@
+Propsal:
+Build a small Update Agent leverage pldm-lib/pldm-common to run ontop of OpenBMC on AST2700.
+Use OpenProt as Firmware Device to execute extern staging flow listed below.  
+Update agent will send GetFirmParameters for Inventory and ActivatePendingComponentImage to signal to OpenProt to begin this flow.
```
rusty1968, line 4: Let's state explicitly that ActivatePendingComponentImage is not a DSP0267 command. The standard PLDM T5 activation command is ActivateFirmware (opcode 0x1a).

docs/demo-proposal-flow.md
CourtneyDrant: ActivatePendingComponentImage is the notification.

docs/demo-proposal-flow.md
rusty1968, line 12: We need to add an explicit causal link between the staging firmware and the GetFirmwareParameters command.

docs/demo-proposal-flow.md
```diff
@@ -0,0 +1,47 @@
...
+        OpenProt-->>BMC: Descriptors
+    end
+
+    BMC->>OpenProt: ActivatePendingComponentImage(AST2070 Component Identifier)
+    OpenProt-->>BMC: EstimatedTimeForActivation
+
+    Note over OpenProt,BMC: OpenProt disables access by BMC (notify BMC to shutdown, then pull power)
+
+    OpenProt->>fwspi: Claim mastership
+    OpenProt->>OpenProt: Verify BMC image in staging area
```
rusty1968, line 26: OpenPRoT must persist the outcome of the verification  process across the power cycle.  UA will poll for it later when it comes back up.

docs/demo-proposal-flow.md
FerralCoder: the full flow is out of our scope, but yes I agree in general.
The real full flow would start with a human somewhere at the company running the server - they select the update and "put it into their infrastructure for deployment". That can mean a lot of different things and is not relevant to OpenPRoT.

But where it does become relevant, including to the demo flow, is that the update eventually lands with an update agent running on the BMC. How it gets to that update agent in general is outside our scope and unknown to OpenPRoT

For the sake of the demo, it could be a user running a script from the OpenBMC's console, directed at a particular update image

docs/demo-proposal-flow.md
FerralCoder: actually, it would be the update agent on BMC that needs to know how to update the staging area and coordinate the PLDM flow alongside that.  Some external agent would provide the image to the OpenBMC update image, but we don't care about that part.

So yes, we need to show the update agent performing these steps either before or during the PLDM steps.

docs/demo-proposal-flow.md
FerralCoder: some breadcrumbs should be put in place to help the system recover properly in the event of a power failure, but in the normal case, the external PRoT itself would not be power cycled here, just the BMC being updated.

docs/demo-proposal-flow.md
FerralCoder: is it possible instead for OpenBMC to query for the status after it reboots following the update attempt? Pass or fail it will have been reset and needs to complete the process later.

Does that fit with how PLDM works, or is there a concept of a "session" that has been destroyed due to the reboot?

docs/demo-proposal-flow.md
rusty1968: @FerralCoder , someone needs to start the flow - via a script , something - it won't start on its own. You may say it is not the real flow, but an external trigger must exist.

FerralCoder: @rusty1968 yes i agree, thats what the last line in my comment was meant to be "For the sake of the demo, it could be a user running a script from the OpenBMC's console, directed at a particular update image"

docs/demo-proposal-flow.md
CourtneyDrant: That is incorrect.  It is a standard DSP0267 command.

docs/demo-proposal-flow.md
chrysh: @FerralCoder Maybe @rusty1968  is referring to the self-update case?

docs/demo-proposal-flow.md
hemanthkumarm-ami, line 4: This approach seems to move away from the standard PLDM Type 5 update flow. One concern is determining how the Update Agent would know when to issue ActivatePendingComponentImage() without introducing behavior that impacts existing component update flows.

To keep the solution generic and fully backward compatible, could we instead introduce an OEM command such as OEM:SetImageSource, which provides the firmware device with the image offset, image length, hash, and any other required metadata?

**Proposed behavior:**

If the firmware device supports OEM:SetImageSource, the Update Agent sends the image location information and then directly issues ActivatePendingComponentImage().
During activation, the firmware device can claim SPI ownership, retrieve the image from the specified location, perform verification, program the target flash, and reset the BMC if needed.
If the command is not supported, the Update Agent falls back to the existing PLDM Type 5 flow (UpdateComponent → RequestFirmwareData → TransferComplete → VerifyComplete → ApplyComplete → ActivatePendingComponentImage).

This keeps the standard Type 5 workflow unchanged for all existing devices while providing a generic extension mechanism for devices that need deferred image retrieval.
````text
UA                               FD
|                                |
| GetFirmwareParameters          |
|------------------------------->|
|<-------------------------------|
|
| RequestUpdate                  |
|------------------------------->|
|<-------------------------------|
|
| PassComponentTable             |
| (CompID, Version, Size)        |
|------------------------------->|
|<-------------------------------|
|
| OEM:SetImageSource             |
| (Offset, Length, Hash)         |
|------------------------------->|
|
+----------------------------------------------+
| Supported ?                                  |
+----------------------------------------------+
        | Yes                    | No
        v                        v

  ActivatePending          Standard Type 5 Flow
  ComponentImage           (UpdateComponent ->
        |                  RequestFirmwareData ->
        |                  TransferComplete ->
        |                  VerifyComplete ->
        |                  ApplyComplete ->
        |                  ActivatePendingComponentImage)
        |
        v

  Claim SPI Ownership
          |
          v
  Locate Image (Offset, Length)
          |
          v
  Verify Hash / Signature
          |
          v
  Program Inactive Partition
          |
          v
  Verify Programmed Image
          |
          v
  Switch Active Partition
          |
          v
  Reboot / Reset BMC
````

CourtneyDrant: We understand and have discussed this.  The flow needs to be updated to include the discover of FD commands by the UA to FD.  We can use GetPLDMCommands to find ActivatePendingComponentImage and GetFirmwareParameters for ComponentActivationMethods.
The pldm-lib has callbacks for a FD to execute upon seeing an ActivatePendingComponent image.  Our plan was to use those callbacks for dealing with the specific component received in the message.

ramabisa1:
ActivatePendingComponentImage should be sent only when FD or FDP is in the IDLE state.  From your proposal, do you think FD (PRoT) is in the IDLE state when ActivatePendingComponentImage command is issued from BMC?

CourtneyDrant:
> ActivatePendingComponentImage should be sent only when FD or FDP is in the IDLE state. From your proposal, do you think FD (PRoT) is in the IDLE state when ActivatePendingComponentImage command is issued from BMC?

Yes, the FD is in Idle until it receives a RequestUpdate.  In this flow, there are no RequestUpdates.  If in the future we want to support RequestUpdate and ActivatePendingComponentImage, we would deny the ActivatePending if already in an RequestUpdate flow.  The UA would then need to retry.

docs/demo-proposal-flow.md
CourtneyDrant: I'll update the Inventory section to specify that GetFirmwareParameters needs to return support for ComponentActivationMethods.ActivatePendingImage.  Self-contained activation, required for ActivatePendingComponentImage, can use an "ActiveFirmware" to activate but it isn't required.  Please see the quotes from the 1.3 spec below.

"12.16 ActivatePendingComponentImage Command Format
requested component 1565 image must support self-contained activation."

"3.26 self-contained activation
In some cases a firmware component is not actively running (i.e., a UEFI driver which only executes on system startup) and therefore the self-contained activation will still apply."

docs/demo-proposal-flow.md
gmandyam: a) What is BMC image verified against?  Maybe mention image is verified against pre-provisioned component ID/RIM.
b) Can verification result be retained?

docs/demo-proposal-flow.md
CourtneyDrant: @FerralCoder I agree with Anthony.  How do we expect to stash the image in the staging area before we trigger the out-of-transport flow?

hemanthkumarm-ami:
The current flow diagram doesn't show the staging action anywhere. Can we add the staging step to the flow? the UA only knows out-of-transport activation is supported once it queries GetFirmwareParameters and sees ComponentActivationMethods.ActivatePendingImage, so ideally it stages the image after confirming that method is available, then calls ActivatePendingComponentImage. is my understanding correct ?

CourtneyDrant:
> The current flow diagram doesn't show the staging action anywhere. Can we add the staging step to the flow? the UA only knows out-of-transport activation is supported once it queries GetFirmwareParameters and sees ComponentActivationMethods.ActivatePendingImage, so ideally it stages the image after confirming that method is available, then calls ActivatePendingComponentImage. is my understanding correct ?

Agreed and this was mentioned above.  @FerralCoder and @rusty1968 Can you help?

PR #457, opened 2026-09-03
pldm: Add the pldm-notify IPC channel (PoC)

chrysh:
## Summary

- Wire protocol (`services/pldm/notify-api`) for PLDM-to-orchestrator notifications. One opcode today (`UpdateRequested`), two responses (`Accepted`, `Rejected`) so the orchestrator can veto before the FD accepts the UA's request.
- Orchestrator adapter (`services/orchestrator/adapters/pldm`) translates wire vocabulary to orchestrator-side `Notification` and encodes the caller's accept/reject verdict.
- QEMU integration test (`target/ast10x0/tests/pldm_notify`) wires two pw_kernel processes on a real channel, always accepting for the PoC.
- Host test counter in `firmware_update_host.rs` tracks how many times the FD fires a notification callback.

Header is `[op/code:1B][len:1B][reserved:2B]` each way, 4 bytes per frame. Both enums are `#[non_exhaustive]` (the two sides are separate processes). The opcode space is flat, not update-specific.

The PoC always accepts. Veto logic, starvation mitigation, and PLDM FD integration come later.

## Test plan

- [ ] `bazel test //services/pldm/notify-api:pldm_notify_api_test`
- [ ] `bazel test //services/orchestrator/adapters/pldm:orchestrator_pldm_adapter_test`
- [ ] `bazel test //services/pldm:firmware_update_host_test`
- [ ] `bazel test --config=virt_ast10x0 //target/ast10x0/tests/pldm_notify:pldm_notify_qemu_test`

Stays a draft until #458 (the PLDM/orchestrator IPC design doc) is merged.

Assisted-by: Claude

### DIFF
--- a/services/orchestrator/adapters/pldm/src/lib.rs
+++ b/services/orchestrator/adapters/pldm/src/lib.rs
@@ -1,88 +1,132 @@
 // Licensed under the Apache-2.0 license
 // SPDX-License-Identifier: Apache-2.0
 
-//! PLDM-backed adapter for the Boot Orchestrator's update-request input.
+//! Orchestrator side of the pldm-notify wire: decodes one notification from the
+//! PLDM service and says what it means in the orchestrator's terms.
 //!
-//! [`UpdateRequestLatch`] binds the PLDM firmware-device service's
-//! [`FdEventSink`] seam to the orchestrator's
-//! [`Event::UpdateRequest`]: the PLDM run loop notifies the latch when the
-//! Update Agent's `RequestUpdate` is accepted, and the orchestrator run loop
-//! drains it with [`take`](UpdateRequestLatch::take). This crate depends on
-//! both stacks by design — the PLDM service stays orchestrator-free and the
-//! orchestrator stays transport-free, the same rule that keeps HAL adapters
-//! out of `orchestrator-capabilities`.
+//! Nothing here reaches the state machine. The notification arrives before the
+//! FD has accepted the UA's request, so there is nothing to authenticate yet;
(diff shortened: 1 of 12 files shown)


services/orchestrator/adapters/pldm/src/lib.rs
```diff
@@ -1,88 +1,132 @@
...
+/// vocabulary across a process boundary, and letting it reach the run loop is
+/// the coupling the adapter exists to prevent.
+///
+/// [`NotifyOp`]: openprot_pldm_notify_api::NotifyOp
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+#[non_exhaustive]
+pub enum Notification {
+    /// A UA sent a `RequestUpdate`. The orchestrator's verdict decides whether
+    /// the FD accepts; no firmware bytes have moved yet.
+    UpdateRequested,
```
CourtneyDrant: We are looking at two different types of Updates from Pldm.  The first is a RequestUpdate received by the UpdateAgent.  This invokes the FD FSM to begin Component Image Transfer of components and the storage into staging memory provided to the PLDM Service.  The other being ActivatePendingComponentImage which we intend to use to initiate component download out of transport (ex: SPI).  
Should we distinguish between the two with this enum?

chrysh:
Superseded by the design in #464. The notify-api wire protocol here (flat opcode + accept/reject) is replaced by the nudge + QueryStatus + grant model. The wire crate and adapter don't carry over; the QEMU two-process scaffold might be salvageable later. Closing.

PR #458, opened 2026-09-07
docs: PLDM/orchestrator IPC design

chrysh:
## Summary

- PLDM/orchestrator IPC design as a sequence diagram, plus a write-access containment section
- Two kernel channels (notify + intake), both initiated by PLDM
- Firmware bytes stay out of IPC (direct to flash, PLDM tracks progress locally)
- Write access is confined to the staging region for the duration of the transfer: typed StagingWindow in PLDM, backed by an SMC write filter the orchestrator controls
- Verified: USER signal latches (level-triggered), MCTP server buffers 4 messages, poll_stage bounded by flash erase time

Related project board items (https://github.com/orgs/9elements/projects/44):
- 9elements/openprot#8 (Update of single downstream device firmware)
- 9elements/openprot#11 (dep: PLDM-T5)

The settled design decisions and the open questions live in the page itself. Two questions are still open: who owns the SPI flash controller, and how the orchestrator learns that PLDM died short of the timeout.

```mermaid
sequenceDiagram
    participant UA as UA (BMC)<br/>remote, over MCTP
    participant PLDM as PLDM FirmwareDevice<br/>single thread: run_terminus
    participant Orch as Orchestrator<br/>single thread: object_wait loop
    participant Flash as Shared Storage<br/>ext. SPI flash

    Note over UA, Orch: NOTIFY CHANNEL (pre-transfer veto)

    UA->>PLDM: RequestUpdate (MCTP)
    activate PLDM
    PLDM->>Orch: channel_transact: Request::UpdateRequested
    Note right of Orch: check state, policy
    Orch-->>PLDM: Response::Accepted | Rejected
    deactivate PLDM
    PLDM-->>UA: RequestUpdate response (accept/reject)

    Note over UA, Orch: if Accepted: INTAKE CHANNEL

    activate PLDM
    PLDM->>Orch: Offer { target: TargetId, total: u64 }
    Note right of Orch: validate target + length,<br/>reserve staging,<br/>open the SMC write filter
    Orch-->>PLDM: IntakeStatus::Receiving { base: FlashAddress, total }
    deactivate PLDM

    loop FD pulls chunks from UA via RequestFirmwareData
        PLDM->>UA: RequestFirmwareData (MCTP)
        UA-->>PLDM: firmware chunk response
        PLDM-->>Flash: write firmware bytes (direct, no IPC)
        Note right of PLDM: PLDM tracks its own write progress
    end

    activate PLDM
    PLDM->>Orch: Complete { written: u64 }
    Note right of Orch: check coverage,<br/>queue Pending::UpdateRequest
    Orch-->>PLDM: IntakeStatus::Authenticating
    deactivate PLDM

    Note over UA, Flash: async: orchestrator event loop drains pending

    PLDM->>UA: TransferComplete (MCTP)

    Note over PLDM: PLDM FREE:<br/>services UA on MCTP<br/>MCTP responsive

    Note over Orch, Flash: EFFECT CHAIN (non-blocking steps)<br/>1. poll_pending<br/>2. SM: Ready -> Updating<br/>3. poll_stage (one step)<br/>4. return to object_wait<br/>repeat 3-4 until phase done<br/>IPC responsive between steps

    Orch-->>Flash: PayloadSource::read_at
    Flash-->>Orch: payload bytes

    Orch->>PLDM: object_set_peer_user_signal<br/>(dataless nudge, wakes WaitGroup)

    loop wake on USER signal, poll status, send *Complete to UA
        activate PLDM
        PLDM->>Orch: Poll
        Note right of Orch: read latched IntakeStatus
        Orch-->>PLDM: Authenticating | Staging | Staged | Failed
        deactivate PLDM
        Note over PLDM, UA: when phase done:
        PLDM->>UA: VerifyComplete (MCTP)
        PLDM->>UA: ApplyComplete (MCTP)
        Note right of PLDM: on failure: same commands<br/>with error completion code
    end

    UA->>PLDM: ActivateFirmware (MCTP, explicit)
    activate PLDM
    PLDM->>Orch: Activate
    Orch-->>PLDM: IntakeStatus::Activating
    deactivate PLDM
    PLDM-->>UA: ActivateFirmware response (accepted, not done)
    Note right of Orch: activation effect (async):<br/>bump SVN in OTP (irreversible),<br/>nudge + Poll reports Activated
    UA->>PLDM: GetStatus (MCTP, until activation lands)
    PLDM-->>UA: current state + AuxState

    Note over UA, Orch: between Offer and Activate
    UA->>PLDM: CancelUpdate (MCTP)
    activate PLDM
    PLDM->>Orch: Abort
    Note right of Orch: in-flight flash step completes<br/>and is discarded
    Orch-->>PLDM: IntakeStatus::Idle
    deactivate PLDM
    PLDM-->>UA: CancelUpdate response

    Note over Orch: If PLDM dies mid-transfer (no Complete, no Abort),<br/>orchestrator-side timeout releases the staging reservation.<br/>The transfer loop is zero-IPC, so this bounds total transfer time:<br/>it must exceed worst-case transfer plus FD_T1 (120s),<br/>so a live PLDM always aborts first.

    Note over UA, Flash: Blocking direction: always PLDM -> Orchestrator, never the reverse.<br/>Every IPC response is immediate. Effects run async via poll_stage (one step, return, repeat).<br/>PLDM stays free to service UA on MCTP. USER signal nudge replaces blind polling.<br/>FD initiates TransferComplete, VerifyComplete, ApplyComplete. UA initiates ActivateFirmware, GetStatus and CancelUpdate.
```

## Test plan

- [ ] Review the diagram and the write-access containment section in `docs/src/design/orchestrator/pldm-orchestrator-ipc.md`
- [ ] Comment on the two open questions

Assisted-by: Claude

### DIFF
--- a/docs/src/design/orchestrator/pldm-orchestrator-ipc.md
+++ b/docs/src/design/orchestrator/pldm-orchestrator-ipc.md
@@ -23,6 +23,13 @@ Design decisions:
 - A Rejected veto becomes an error completion code in the RequestUpdate
   response, ALREADY_IN_UPDATE_MODE when the reason is an update already
   running; the UA retries.
+- Write access is two layers: a typed StagingWindow inside PLDM (Rust, catches
+  offset bugs) backed by an SMC write filter PLDM cannot reprogram (catches a
+  compromised process). The orchestrator opens the filter on Offer and closes
+  it on Complete/Abort/timeout. How the filter registers are kept out of
+  PLDM's reach (separate MPU region, separate controller/CS, lock-until-reset)
+  depends on the AST10x0 register layout; see "Who owns the SPI flash
+  controller" in open questions.
 - Transfer loop is zero-IPC: PLDM writes firmware bytes direct to flash and
   tracks progress locally. Complete carries the byte count for the
   orchestrator's coverage check (early-fail only, verify hashes the staged
@@ -116,6 +123,39 @@ sequenceDiagram
     Note over UA, Flash: Blocking direction: always PLDM -> Orchestrator, never the reverse.<br/>Every IPC response is immediate. Effects run async via poll_stage (one step, return, repeat).<br/>PLDM stays free to service UA on MCTP. USER signal nudge replaces blind polling.<br/>FD initiates TransferComplete, VerifyComplete, ApplyComplete. UA initiates ActivateFirmware, GetStatus and CancelUpdate.
 ```
 
+## Write-access containment
+
+PLDM writes firmware bytes direct to flash (zero-IPC, see above), so write
+access must be confined to the inactive slot and only for the duration of the
+transfer. Two layers, each catching a different class of failure:
+
+The first layer is a typed StagingWindow inside the PLDM process. When PLDM
+receives a Receiving response it constructs the window: a bounded handle over
+the inactive slot (base address + length, capped to slot size). All writes go
+through the window; it translates offsets and rejects anything outside the
+region. The window is dropped on Complete, Abort, or timeout, so PLDM holds
+no flash handle outside an active transfer. This catches offset bugs and
+use-after-transfer bugs but not a compromised process, because PLDM still has
+the underlying flash mapped. Whether Receiving carries an explicit base or the
+staging region is board-static is an open question (see below).
+
+The second layer is a hardware write filter that PLDM cannot reprogram. The
+SMC raises SmcInterrupt::WriteProtected on writes outside an allowed region
+(interrupts.rs:59). The orchestrator (or a dedicated flash-service process)
+opens the filter for the staging region on Offer and closes it on
+Complete/Abort/timeout. PLDM needs the SMC control registers that drive
+erase/program commands, but must not be able to touch the filter/write-protect
+registers. Whether those register sets are separable (distinct MPU pages,
+separate controller/CS, or lock-until-reset bits) depends on the AST10x0
+register layout and is folded into the "who owns the SPI flash controller"
+open question below.
+
+The net effect: bugs hit the Rust window check, a compromised process hits the
+hardware filter, and both "inactive slot only" and "only during an update" are
+enforced. Even a fully rogue PLDM can at worst corrupt the staging area and
+fail verify; the active image is never written by PLDM at any point in the
+flow, and activation is orchestrator-side metadata plus the SVN bump in OTP.
+
 ## Open questions
 
 Who owns the SPI flash controller. The diagram has PLDM writing the staging
@@ -124,7 +164,16 @@ says nothing about multiple clients. Either each process drives its own
 controller over disjoint regions, or one process owns the driver and the other
 reaches it over IPC. Sequencing keeps the two off the same bytes at the same
 time (the orchestrator reads only after Complete), so this is about the driver
-and the controller, not about the protocol.
+and the controller, not about the protocol. A related constraint from the
+write-access containment section: PLDM needs the erase/program control
+registers but must not reach the write-protect/filter registers. Whether those
+register sets fall on separate MPU pages on the AST10x0 (datasheet needed)
+determines whether pw_kernel can enforce the split, or whether a dedicated
+flash-service process must own the entire SMC and proxy writes.
+
+Whether Receiving carries an explicit base address for the staging region or
+the region is board-static. The typed StagingWindow needs a base; today
+Receiving only carries total.
 
 Whether the kernel can tell the orchestrator that PLDM's channel closed. That
 would replace the transfer-time timeout, which has to be generous.
--- a/docs/src/design/orchestrator/pldm-orchestrator-ipc.md
+++ b/docs/src/design/orchestrator/pldm-orchestrator-ipc.md
@@ -1,0 +1,110 @@
+# PLDM/Orchestrator IPC
+
+How the PLDM FirmwareDevice service and the orchestrator communicate during a
+firmware update. Two Pigweed kernel channels, both initiated by PLDM: a notify
+channel for pre-transfer veto, and an intake channel for control messages
+(Offer, Complete, Poll, Activate, Abort) and the async effect chain. Firmware
+bytes go direct to flash, never through IPC.
+
+Design decisions:
+
+- Activate is on the wire (ActivateFirmware from the UA), not implicit after
+  staging.
+- Flash seam is async: poll_stage calls start_erase/start_program, returns,
+  checks is_busy on the next call. Uses FlashDriver's split API, not
+  BlockingFlash.
+- USER signal is level-triggered (verified from Pigweed kernel source): OR'd
+  into the peer's active_signals bitfield, persists until lowered. No lost
+  wakeups.
+- MCTP server (separate process) buffers 4 messages while PLDM is in a
+  transact. Overflow drops silently, no backpressure to the bus.
+- Transfer loop is zero-IPC: PLDM writes firmware bytes direct to flash and
+  tracks progress locally. Complete carries the byte count for the
+  orchestrator's coverage check (early-fail only, verify hashes the staged
+  image anyway). On a write error PLDM can nudge and report Failed via Poll,
+  no extra IPC verb needed.
+
+```mermaid
+sequenceDiagram
+    participant UA as UA (BMC)<br/>remote, over MCTP
+    participant PLDM as PLDM FirmwareDevice<br/>single thread: run_terminus
+    participant Orch as Orchestrator<br/>single thread: object_wait loop
+    participant Flash as Shared Storage<br/>ext. SPI flash
+
+    Note over UA, Orch: NOTIFY CHANNEL (pre-transfer veto)
+
+    UA->>PLDM: RequestUpdate (MCTP)
+    activate PLDM
+    PLDM->>Orch: channel_transact: Request::UpdateRequested
+    Note right of Orch: check state, policy
+    Orch-->>PLDM: Response::Accepted | Rejected
+    deactivate PLDM
+    PLDM-->>UA: RequestUpdate response (accept/reject)
+
+    Note over UA, Orch: if Accepted: INTAKE CHANNEL
+
+    activate PLDM
+    PLDM->>Orch: Offer { target: TargetId, total: u64 }
+    Note right of Orch: validate target + length,<br/>reserve staging
+    Orch-->>PLDM: IntakeStatus::Receiving { total }
+    deactivate PLDM
+
+    loop FD pulls chunks from UA via RequestFirmwareData
+        PLDM->>UA: RequestFirmwareData (MCTP)
+        UA-->>PLDM: firmware chunk response
+        PLDM-->>Flash: write firmware bytes (direct, no IPC)
+        Note right of PLDM: PLDM tracks its own write progress
+    end
+
+    activate PLDM
+    PLDM->>Orch: Complete { written: u64 }
+    Note right of Orch: check coverage,<br/>queue Pending::UpdateRequest
+    Orch-->>PLDM: IntakeStatus::Authenticating
+    deactivate PLDM
+
+    Note over UA, Flash: async: orchestrator event loop drains pending
+
+    PLDM->>UA: TransferComplete (MCTP)
+
+    Note over PLDM: PLDM FREE:<br/>services UA on MCTP<br/>MCTP responsive
+
+    Note over Orch, Flash: EFFECT CHAIN (non-blocking steps)<br/>1. poll_pending<br/>2. SM: Ready -> Updating<br/>3. poll_stage (one step)<br/>4. return to object_wait<br/>repeat 3-4 until phase done<br/>IPC responsive between steps
+
+    Orch-->>Flash: PayloadSource::read_at
+    Flash-->>Orch: payload bytes
+
+    Orch->>PLDM: object_set_peer_user_signal<br/>(dataless nudge, wakes WaitGroup)
+
+    loop wake on USER signal, poll status, send *Complete to UA
+        activate PLDM
+        PLDM->>Orch: Poll
+        Note right of Orch: read latched IntakeStatus
+        Orch-->>PLDM: Authenticating | Staging | Staged | Failed
+        deactivate PLDM
+        Note over PLDM, UA: when phase done:
+        PLDM->>UA: VerifyComplete (MCTP)
+        PLDM->>UA: ApplyComplete (MCTP)
+        Note right of PLDM: on failure: same commands<br/>with error completion code
+    end
+
+    UA->>PLDM: ActivateFirmware (MCTP, explicit)
+    activate PLDM
+    PLDM->>Orch: Activate
+    Orch-->>PLDM: IntakeStatus::Activating
+    deactivate PLDM
+    PLDM-->>UA: ActivateFirmware response
+    Note right of Orch: activation effect (async):<br/>bump SVN in OTP (irreversible),<br/>nudge + Poll reports Activated
+
+    Note over UA, Orch: between Offer and Activate
+    UA->>PLDM: CancelUpdate (MCTP, 0x1D)
+    activate PLDM
+    PLDM->>Orch: Abort
+    Note right of Orch: in-flight flash step completes<br/>and is discarded
+    Orch-->>PLDM: IntakeStatus::Idle
+    deactivate PLDM
+    PLDM-->>UA: CancelUpdate response
+
+    Note over Orch: If PLDM dies mid-transfer (no Complete, no Abort),<br/>orchestrator-side timeout releases the staging reservation.
+
+    Note over UA, Flash: Blocking direction: always PLDM -> Orchestrator, never the reverse.<br/>Every IPC response is immediate. Effects run async via poll_stage (one step, return, repeat).<br/>PLDM stays free to service UA on MCTP. USER signal nudge replaces blind polling.<br/>FD initiates TransferComplete, VerifyComplete, ApplyComplete. UA initiates ActivateFirmware.
+```
--- a/docs/src/design/orchestrator/pldm-orchestrator-ipc.md
+++ b/docs/src/design/orchestrator/pldm-orchestrator-ipc.md
@@ -11,13 +11,18 @@ Design decisions:
 - Activate is on the wire (ActivateFirmware from the UA), not implicit after
   staging.
 - Flash seam is async: poll_stage calls start_erase/start_program, returns,
-  checks is_busy on the next call. Uses FlashDriver's split API, not
-  BlockingFlash.
+  checks is_busy on the next call and then complete_op, which is where the
+  operation's error surfaces. Uses FlashDriver's split API, not BlockingFlash.
 - USER signal is level-triggered (verified from Pigweed kernel source): OR'd
   into the peer's active_signals bitfield, persists until lowered. No lost
   wakeups.
 - MCTP server (separate process) buffers 4 messages while PLDM is in a
-  transact. Overflow drops silently, no backpressure to the bus.
+  transact. Overflow drops silently, no backpressure to the bus. Recovery from
+  a dropped message is PLDM's, not this seam's: pldm-lib defaults FD_T1 (update
+  mode idle) to 120s and FD_T2 (RequestFirmwareData retry) to 5s.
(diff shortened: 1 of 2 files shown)


chrysh:
Interactive version of the sequence diagram (self-contained HTML, download and open in a browser):
https://github.com/9elements/openprot/raw/worktree-docs-pldm-ipc/docs/src/design/orchestrator/pldm-orchestrator-ipc.html

chrysh:
My open questions (mostly to myself) after talking with @embediver, and me answering them myself:

1. Does the PLDM process have direct write access to the flash? What if it is a malicious PLDM service?
> The SPIPF filter watches the host, not the RoT's own writes. If PLDM code is compromised, the software-only Updatable trait contract is all that stands between it and the active slot.

2. How/where is the async Verification thread spawned?
> No async thread. In pw_kernel, you need to define the processes at compile time, they cannot be spawned dynamically. Therefore, we decided for the IncrementalVerifier setup, where the orchestrator verifies one chunk per loop, so that it does not block.

3. How does the SPI Monitor influence the write and boot process?
>   The SPIPF hardware sits between the host SPI master and the flash chip. It does three       
  things:

  a. Mux control (MuxSelect): RoT takes exclusive flash access during boot/verify (RotControl), then hands back to host (HostControl)
  b. Address privilege filtering: 256 MiB space divided into 16 KB blocks, each independently allow/deny for read and write. The RoT configures RegionPolicy::deny(start, length, Write) over the active slot, then calls lock() to make the policy immutable until reset
  c. Command filtering: allow-list of SPI opcodes, so write-enabling commands (WREN, page program, sector erase) can be excluded entirely

  Boot sequence: Start → Hold (RoT exclusive) → ConfigurePolicy → Release (host) → RuntimeMonitoring

  Critical limitation: this filters the host's access. The RoT's own writes (through FMC or SPI1/SPI2) bypass the monitor entirely. That's why the malicious-PLDM question matters.

4. Do we have just one flash chip, e.g. with one active slot region and the inactive region? How do we write protect the active slot/memory region? SPI Monitor?
>   Slot layout is deliberately abstract in the codebase. The concrete layout (addresses, sizes, how many chips) lives in board-specific code that doesn't exist yet. The hardware supports multiple flash chips via FMC (CE0/CE1) and SPI1/SPI2 controllers. 
Write protection of the active slot comes from the SPIPF address privilege bitmap: deny writes to the active region, lock the policy. Whether that's one chip with two regions or two chips with separate chip-selects is a board decision.

docs/src/design/orchestrator/pldm-orchestrator-ipc.md
```diff
@@ -0,0 +1,193 @@
...
+
+How the PLDM FirmwareDevice service and the orchestrator communicate during a
+firmware update. Two Pigweed kernel channels, both initiated by PLDM: a notify
+channel for pre-transfer veto, and an intake channel for control messages
+(Offer, Complete, Poll, Activate, Abort) and the async effect chain. Firmware
+bytes go direct to flash, never through IPC.
+
+Design decisions:
+
+- Activate is on the wire (ActivateFirmware from the UA), not implicit after
```
CourtneyDrant: We may need to mention that we require "SelfContainedActivationRequest" and "Self-Contained" in FirmwareParameters.  No other activations are supported.

docs/src/design/orchestrator/pldm-orchestrator-ipc.md
```diff
@@ -0,0 +1,193 @@
+# PLDM/Orchestrator IPC
+
+How the PLDM FirmwareDevice service and the orchestrator communicate during a
+firmware update. Two Pigweed kernel channels, both initiated by PLDM: a notify
```
rusty1968, line 4: The Orchestrator is a supervisor of the update scenario. It is not a channel handler - it is the initiator. It will register for notifications from PLDM when it is ready to do so. Use the I2C/MCTP as a template (with the bounded wait of course which is already part of the skeleton I have laid out). PLDM should be the server/handler, and the orchestrator the client that gets notified.

Look at what we already do today - The i2c server (handler) nudges its client via object_set_peer_user_signal on an async event, and the client then issues a transact to pull the latched data.

chrysh: In the flipped model, how does the Accepted/Rejected verdict reach PLDM before the RequestUpdate response goes back to the UA, a second transact?

channel_transact is an initiator-only. In the flipped model you are suggesting, every exchange where PLDM needs an answer (veto, Offer, Complete, Activate, Abort) becomes: latch → nudge → orchestrator pulls event → orchestrator pushes verdict in a second transact. Two round trips per exchange, sync or async, while PLDM holds the UA's request pending. Current design: one transact.

It would make the status-push direction cleaner (Authenticating/Staging/Staged), but I am not even sure how important this information is for PLDM.

@CourtneyDrant What is your opinion?

chrysh: @rusty1968 In the case of I2C, the client only pulls latched bytes. In our case, it's PLDM that needs the answer, and should be the one initiating the interaction, because PLDM needs to answer the UA. PLDM would need to hold UA state across: latch, nudge, orchestrator pulls, orchestrator pushes verdict in a second transact.

Furthermore, it moves blocking into the orchestrator's loop. channel_transact blocks the caller. I wanted the orchestrator to be always responsive.

That being said, if we as a group think the PLDM being the server is the better software design, I am happy to rewrite it. But given that the request direction is always UA -> FD, it seems to me that pldm being the client asking the orchestrator feels like the more natural direction.

docs/src/design/orchestrator/pldm-orchestrator-ipc.md
```diff
@@ -0,0 +1,193 @@
...
+- Transfer loop is zero-IPC: PLDM writes firmware bytes direct to flash and
+  tracks progress locally. Complete carries the byte count for the
+  orchestrator's coverage check (early-fail only, verify hashes the staged
+  image anyway). On a write error PLDM sends Abort to release staging and
+  reports the failure to the UA in TransferComplete's result code.
+
+```mermaid
+sequenceDiagram
+    participant UA as UA (BMC)<br/>remote, over MCTP
+    participant PLDM as PLDM FirmwareDevice<br/>single thread: run_terminus
```
rusty1968, line 54: I am thinking we should rename pldm as fw_update service. PLDM as a middleware crate is distinct from the service that uses it.

docs/src/design/orchestrator/pldm-orchestrator-ipc.md
```diff
@@ -0,0 +1,193 @@
...
+  the window PLDM writes through and the window the hardware allows come from
+  one place. PLDM holds no board layout.
+- Write access is two layers: a typed StagingWindow inside PLDM (Rust, catches
+  offset bugs) backed by an SMC write filter PLDM cannot reprogram (catches a
+  compromised process). The orchestrator opens the filter on Offer and closes
+  it on Complete/Abort/timeout. How the filter registers are kept out of
+  PLDM's reach (separate MPU region, separate controller/CS, lock-until-reset)
+  depends on the AST10x0 register layout; see "Who owns the SPI flash
+  controller" in open questions.
+- Transfer loop is zero-IPC: PLDM writes firmware bytes direct to flash and
```
rusty1968, line 45: There is an IPC channel between the fw_update service and the flash driver binding that manages the staging partition. Perhaps the more appropriate label should be zero Orchestrator IPC?

docs/src/design/orchestrator/pldm-orchestrator-ipc.md
CourtneyDrant: > In the flipped model, how does the Accepted/Rejected verdict reach PLDM before the RequestUpdate response goes back to the UA, a second transact?
> 
> channel_transact is an initiator-only. In the flipped model you are suggesting, every exchange where PLDM needs an answer (veto, Offer, Complete, Activate, Abort) becomes: latch → nudge → orchestrator pulls event → orchestrator pushes verdict in a second transact. Two round trips per exchange, sync or async, while PLDM holds the UA's request pending. Current design: one transact.
> 
> It would make the status-push direction cleaner (Authenticating/Staging/Staged), but I am not even sure how important this information is for PLDM.
> 
> [@CourtneyDrant](https://github.com/CourtneyDrant) What is your opinion?

I think a second RequestUpdate would return a AlreadyInUpdateMode by request_update_rsp in fd_context.  The response from the orchestrator is superfluous.

rusty1968:
@CourtneyDrant , check this PR assumes in-transport transfer of bytes.

docs/src/design/orchestrator/pldm-orchestrator-ipc.md
```diff
@@ -0,0 +1,193 @@
...
+firmware update. Two Pigweed kernel channels, both initiated by PLDM: a notify
+channel for pre-transfer veto, and an intake channel for control messages
+(Offer, Complete, Poll, Activate, Abort) and the async effect chain. Firmware
+bytes go direct to flash, never through IPC.
+
+Design decisions:
+
+- Activate is on the wire (ActivateFirmware from the UA), not implicit after
+  staging.
+- Flash seam is async: poll_stage calls start_erase/start_program, returns,
```
rusty1968, line 13: The flash service API is blocking - what you are referring to here is the FlashDriver trait whose implementor resides on the flash service process.

See https://github.com/OpenPRoT/openprot/pull/365

docs/src/design/orchestrator/pldm-orchestrator-ipc.md
```diff
@@ -0,0 +1,193 @@
+# PLDM/Orchestrator IPC
+
+How the PLDM FirmwareDevice service and the orchestrator communicate during a
```
rusty1968, line 3: The firmware_device run_terminus_inner is currrently a tight loop interleaving the roles of initiator/responder - what you are assuming here is that we can fit the blocking requests to the orchestrator in the FdOps callbacks without disrupting the protocol.

chrysh:
Snapshotting the sequence diagram as it stands at 7a63481, so this thread keeps
a fixed version to argue against while the diagram itself changes. The next
revision folds in the FdOps blocking question, the flash service in #365, and
out-of-transport activation from pldm-lib #20.

```mermaid
sequenceDiagram
    participant UA as UA (BMC)<br/>remote, over MCTP
    participant PLDM as PLDM FirmwareDevice<br/>single thread: run_terminus
    participant Orch as Orchestrator<br/>single thread: object_wait loop
    participant Flash as Shared Storage<br/>ext. SPI flash

    Note over UA, Orch: NOTIFY CHANNEL (pre-transfer veto)

    UA->>PLDM: RequestUpdate (MCTP)
    activate PLDM
    PLDM->>Orch: channel_transact: Request::UpdateRequested
    Note right of Orch: check state, policy
    Orch-->>PLDM: Response::Accepted | Rejected
    deactivate PLDM
    PLDM-->>UA: RequestUpdate response (accept/reject)

    Note over UA, Orch: if Accepted: INTAKE CHANNEL

    activate PLDM
    PLDM->>Orch: Offer { target: TargetId, total: u64 }
    Note right of Orch: validate target + length,<br/>reserve staging,<br/>open the SMC write filter
    Orch-->>PLDM: IntakeStatus::Receiving { base: FlashAddress, total }
    deactivate PLDM

    loop FD pulls chunks from UA via RequestFirmwareData
        PLDM->>UA: RequestFirmwareData (MCTP)
        UA-->>PLDM: firmware chunk response
        PLDM-->>Flash: write firmware bytes (direct, no IPC)
        Note right of PLDM: PLDM tracks its own write progress
    end

    activate PLDM
    PLDM->>Orch: Complete { written: u64 }
    Note right of Orch: check coverage,<br/>queue Pending::UpdateRequest
    Orch-->>PLDM: IntakeStatus::Authenticating
    deactivate PLDM

    Note over UA, Flash: async: orchestrator event loop drains pending

    PLDM->>UA: TransferComplete (MCTP)

    Note over PLDM: PLDM FREE:<br/>services UA on MCTP<br/>MCTP responsive

    Note over Orch, Flash: EFFECT CHAIN (non-blocking steps)<br/>1. poll_pending<br/>2. SM: Ready -> Updating<br/>3. poll_stage (one step)<br/>4. return to object_wait<br/>repeat 3-4 until phase done<br/>IPC responsive between steps

    Orch-->>Flash: PayloadSource::read_at
    Flash-->>Orch: payload bytes

    Orch->>PLDM: object_set_peer_user_signal<br/>(dataless nudge, wakes WaitGroup)

    loop wake on USER signal, poll status, send *Complete to UA
        activate PLDM
        PLDM->>Orch: Poll
        Note right of Orch: read latched IntakeStatus
        Orch-->>PLDM: Authenticating | Staging | Staged | Failed
        deactivate PLDM
        Note over PLDM, UA: when phase done:
        PLDM->>UA: VerifyComplete (MCTP)
        PLDM->>UA: ApplyComplete (MCTP)
        Note right of PLDM: on failure: same commands<br/>with error completion code
    end

    UA->>PLDM: ActivateFirmware (MCTP, explicit)
    activate PLDM
    PLDM->>Orch: Activate
    Orch-->>PLDM: IntakeStatus::Activating
    deactivate PLDM
    PLDM-->>UA: ActivateFirmware response (accepted, not done)
    Note right of Orch: activation effect (async):<br/>bump SVN in OTP (irreversible),<br/>nudge + Poll reports Activated
    UA->>PLDM: GetStatus (MCTP, until activation lands)
    PLDM-->>UA: current state + AuxState

    Note over UA, Orch: between Offer and Activate
    UA->>PLDM: CancelUpdate (MCTP)
    activate PLDM
    PLDM->>Orch: Abort
    Note right of Orch: in-flight flash step completes<br/>and is discarded
    Orch-->>PLDM: IntakeStatus::Idle
    deactivate PLDM
    PLDM-->>UA: CancelUpdate response

    Note over Orch: If PLDM dies mid-transfer (no Complete, no Abort),<br/>orchestrator-side timeout releases the staging reservation.<br/>The transfer loop is zero-IPC, so this bounds total transfer time:<br/>it must exceed worst-case transfer plus FD_T1 (120s),<br/>so a live PLDM always aborts first.

    Note over UA, Flash: Blocking direction: always PLDM -> Orchestrator, never the reverse.<br/>Every IPC response is immediate. Effects run async via poll_stage (one step, return, repeat).<br/>PLDM stays free to service UA on MCTP. USER signal nudge replaces blind polling.<br/>FD initiates TransferComplete, VerifyComplete, ApplyComplete. UA initiates ActivateFirmware, GetStatus and CancelUpdate.
```

chrysh:
@CourtneyDrant pldm-lib #20 adds ActivatePendingComponent and the
handle_pending_component callback, which is the out-of-transport path Rusty
mentions above. Three questions before I revise the diagram here.

1. Is out-of-transport additive, or the path you expect production to use? If
   it is primary, the transfer loop in this diagram (RequestFirmwareData, the
   staging window, the coverage check on Complete) describes a path nobody
   takes.

2. With out-of-transport, who writes the image into flash before
   ActivatePendingComponent arrives, and how does the RoT learn where it landed
   and how long it is? There is no intake session to carry that, and
   handle_pending_component gets classification, identifier and index only.

3. Does the FD support both entry points at once? Then the orchestrator needs a
   path for an activation with no preceding intake session, where it verifies
   an image it never reserved staging for.

I will draw whichever you confirm.

docs/src/design/orchestrator/pldm-orchestrator-ipc.md
chrysh: Two questions then, because this decides whether the effect chain in this doc
works at all.

Does the flash service own the external SPI flash we stage into, or only
on-chip flash?

And is a split or deferred API on the service planned? `FlashIpcClient` makes
every erase, program and read a blocking transact, and `FlashDriver`'s split
API (`start_erase`/`is_busy`/`complete_op`) stays inside the server process.
Without something equivalent across the boundary, the orchestrator's verifier
blocks for a full chunk read every loop pass and a staging erase blocks it for
10 to 100 ms. That is the "never wait in the handler" property the async effect
chain here is built on.

docs/src/design/orchestrator/pldm-orchestrator-ipc.md
chrysh: The flipped model works cleanly if the veto is allowed to land after the
RequestUpdate response: the orchestrator registers, gets nudged, pulls the
event, and gates whether the transfer starts. It does not work if the verdict
has to be in the response itself, because a handler cannot push to its
initiator, so PLDM would have to hold the UA's request across a nudge and a
second transact.

Which of the two do you have in mind? If it is the first, I will redraw with
the veto as a transfer gate and drop the in-response reject.

docs/src/design/orchestrator/pldm-orchestrator-ipc.md
chrysh: Agreed on the mechanism. `handle_component`, `activate` and
`cancel_update_component` all run inside the responder callback with the UA
waiting, so a transact there holds the UA's response for as long as the
orchestrator takes to answer.

FD_T1 and FD_T2 are ours to set, but the response timeout is the BMC's. Do you
have a number for what the UA tolerates? Your skeleton already carries a
bounded wait, so I assume you had one in mind.

docs/src/design/orchestrator/pldm-orchestrator-ipc.md
rusty1968: > Two questions then, because this decides whether the effect chain in this doc works at all.
> 
> Does the flash service own the external SPI flash we stage into, or only on-chip flash?
> 
> And is a split or deferred API on the service planned? `FlashIpcClient` makes every erase, program and read a blocking transact, and `FlashDriver`'s split API (`start_erase`/`is_busy`/`complete_op`) stays inside the server process. Without something equivalent across the boundary, the orchestrator's verifier blocks for a full chunk read every loop pass and a staging erase blocks it for 10 to 100 ms. That is the "never wait in the handler" property the async effect chain here is built on.

It is one service per flash device, so the answer is yes count on being able to open a channel to the device server managing the spi where the image is stored. The orchestrator is not in the data path though.

chrysh: What do you mean by not in the data path? The orchestrator will tell pldm where to put the firmware though, I assume?

docs/src/design/orchestrator/pldm-orchestrator-ipc.md
rusty1968: > Agreed on the mechanism. `handle_component`, `activate` and `cancel_update_component` all run inside the responder callback with the UA waiting, so a transact there holds the UA's response for as long as the orchestrator takes to answer.
> 
> FD_T1 and FD_T2 are ours to set, but the response timeout is the BMC's. Do you have a number for what the UA tolerates? Your skeleton already carries a bounded wait, so I assume you had one in mind.

The timeouts were not selected by me, let's follow up with @CourtneyDrant .

docs/src/design/orchestrator/pldm-orchestrator-ipc.md
rusty1968: > What do you mean by not in the data path? The orchestrator will tell pldm where to put the firmware though, I assume?

The per-component staging area mapping is platform-specific - by not in the data path I meant to say the orchestrator doesperform any  flash I/O.

docs/src/design/orchestrator/pldm-orchestrator-ipc.md
CourtneyDrant: > In the flipped model, how does the Accepted/Rejected verdict reach PLDM before the RequestUpdate response goes back to the UA, a second transact?
> 
> channel_transact is an initiator-only. In the flipped model you are suggesting, every exchange where PLDM needs an answer (veto, Offer, Complete, Activate, Abort) becomes: latch → nudge → orchestrator pulls event → orchestrator pushes verdict in a second transact. Two round trips per exchange, sync or async, while PLDM holds the UA's request pending. Current design: one transact.
> 
> It would make the status-push direction cleaner (Authenticating/Staging/Staged), but I am not even sure how important this information is for PLDM.
> 
> @CourtneyDrant What is your opinion?

Can we handle the answer asynchronously and respond to subsequent UA initiated messages with an error?  The UA can then send a CancelRequest.

chrysh:
Superseded by #464, which inverts the direction: PLDM-FD is the IPC server and the orchestrator its client, granting or denying each phase. Closing this one.

PR #459, opened 2026-09-08
orchestrator: add IncrementalVerifier capability trait

chrysh:
TL;DR: Polled verification trait for the update pump, consuming-session
design so post-verdict poll is a compile error.

Part of 9elements/openprot#8.

## Summary

Trait-only PR. Adds `IncrementalVerifier` and `VerifySession` to
`orchestrator-capabilities`: polled, chunked image verification so the
update pump never blocks on hashing megabytes of payload.

`start()` consumes the verifier into a `VerifySession`. Each `poll()`
consumes the session: `Processing` returns it inside the variant for the
next call, terminal outcomes return the verifier for reuse. A retained
verdict would be unsafe because `poll()` takes a payload reference per
call, so a swapped payload could get a stale result. The consuming design
makes that unrepresentable.

This diverges from `Updatable` (implicit-start, poll-on-self). Migrating
`Updatable` to the same session shape is a follow-up once its consumers
exist (the pump needs to land first so we can migrate callers).

Driver composition (the adapter that wraps a crypto backend behind this
trait, wired into the update pump) follows in a stacked PR.
`Effect::AuthenticateUpdate` remains stubbed in the SM until then.

Assisted-by: Claude

### DIFF
--- a/services/orchestrator/capabilities/src/incremental_verifier.rs
+++ b/services/orchestrator/capabilities/src/incremental_verifier.rs
@@ -1,0 +1,387 @@
+// Licensed under the Apache-2.0 license
+// SPDX-License-Identifier: Apache-2.0
+
+//! The [`IncrementalVerifier`] update-verification capability contract.
+
+use crate::PayloadSource;
+
+/// Factory for incremental verification sessions. Call [`start`] to
+/// begin hashing a candidate image; the returned [`VerifySession`] does
+/// the actual work one bounded step at a time.
+///
+/// `start` consumes the verifier. Terminal outcomes
+/// ([`Authenticated`](PollOutcome::Authenticated),
+/// [`Rejected`](PollOutcome::Rejected),
+/// [`Fault`](PollOutcome::Fault)) and [`abandon`](VerifySession::abandon)
+/// return it, so the caller can start another session. Polling after a
+/// verdict is unrepresentable: the session is gone.
+///
+/// The boot-time synchronous `Verifier` (in the driver crate) is
+/// unaffected: it stays one-shot for the chain walk, where the image is
+/// small and local.
+///
+/// [`start`]: IncrementalVerifier::start
+pub trait IncrementalVerifier: Sized {
+    /// The error reported when the check itself cannot run: crypto fault
+    /// or unreadable payload. A bad image is
+    /// [`Rejected`](PollOutcome::Rejected), not an error.
+    type Error: core::error::Error;
+
+    /// The session type returned by [`start`](IncrementalVerifier::start).
+    type Session: VerifySession<Verifier = Self, Error = Self::Error>;
+
+    /// Begins a new verification session. The verifier's internal state
+    /// (crypto context, scratch buffers) moves into the session.
+    fn start(self) -> Self::Session;
+}
+
+/// A live verification session. Each [`poll`] call does at most one
+/// read from the payload and one hash update, then returns. The chunk
+/// size is implementor-chosen, sized so each poll fits the caller's
+/// per-poll time budget.
+///
+/// The caller watches progress via `done` in
+/// [`Processing`](PollOutcome::Processing) and abandons a session that
+/// stalls, on its own budget; the session never judges liveness.
+///
+/// `poll` consumes the session and returns a [`PollOutcome`]. On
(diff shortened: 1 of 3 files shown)


services/orchestrator/capabilities/src/incremental_verifier.rs
```diff
@@ -0,0 +1,387 @@
...
+pub trait VerifySession: Sized {
+    /// The verifier this session was created from.
+    type Verifier;
+
+    /// The error type, matching the verifier's.
+    type Error: core::error::Error;
+
+    /// Processes one bounded step. Never waits on device progress, never
+    /// sleeps. An empty payload is a fault, never a vacuous
+    /// `Authenticated`.
```
leongross, line 62:
```suggestion
    /// sleeps. An empty payload is a fault.
```

services/orchestrator/capabilities/src/incremental_verifier.rs
```diff
@@ -0,0 +1,387 @@
...
+    /// Discards the session and returns the verifier.
+    /// Infallible: a half-finished hash state is simply dropped.
+    fn abandon(self) -> Self::Verifier;
+}
+
+/// Result of one [`VerifySession::poll`] call. Terminal variants return
+/// the verifier, ready for a new [`start`](IncrementalVerifier::start).
+#[derive(Debug)]
+pub enum PollOutcome<S: VerifySession> {
+    /// One chunk processed. `done` bytes so far out of `total`.
```
leongross, line 74: Not a fan of `done` here; it implies duality, done or not done. I think we had this somewhere before, did we replace it with `count`?

services/orchestrator/capabilities/src/incremental_verifier.rs
```diff
@@ -0,0 +1,387 @@
+// Licensed under the Apache-2.0 license
+// SPDX-License-Identifier: Apache-2.0
+
+//! The [`IncrementalVerifier`] update-verification capability contract.
+
+use crate::PayloadSource;
+
+/// Factory for incremental verification sessions. Call [`start`] to
+/// begin hashing a candidate image; the returned [`VerifySession`] does
+/// the actual work one bounded step at a time.
```
leongross, line 10: I am confused about the wording. Here you say we hash the images, and later you say we verify the signature? Should this component do both, or is there a mix-up?

services/orchestrator/capabilities/src/incremental_verifier.rs
```diff
@@ -0,0 +1,387 @@
...
+/// Result of one [`VerifySession::poll`] call. Terminal variants return
+/// the verifier, ready for a new [`start`](IncrementalVerifier::start).
+#[derive(Debug)]
+pub enum PollOutcome<S: VerifySession> {
+    /// One chunk processed. `done` bytes so far out of `total`.
+    /// The session is inside, ready for the next poll.
+    Processing { session: S, done: u64, total: u64 },
+    /// The complete image authenticated (signature and policy checks
+    /// passed).
+    Authenticated(S::Verifier),
```
leongross, line 79: In the module description, you say, that this tracks the hashing, which makes sense with the gradual byte supply. How does this couple with signature verification?

services/orchestrator/capabilities/src/incremental_verifier.rs
```diff
@@ -0,0 +1,387 @@
...
+    fn multi_poll_until_authenticated() {
+        let payload = SlicePayload(&[0xAA; 10]);
+        let session = ChunkedVerifier.start();
+
+        // 4 bytes, 4 bytes, 2 bytes = 3 Processing steps, then verdict.
+        let PollOutcome::Processing {
+            session,
+            done: 4,
+            total: 10,
+        } = session.poll(&payload)
```
leongross, line 207: potential future offset drift with hardcoded constant sizes

leongross (commented):
Note from a personal discussion with @chrysh: This only comes into effect if the orchestrator cannot block. Then, we need to incrementally verify one chunk at a time in the main loop.
We will further discuss this architectural decision and conclude whether we merge it as is.

chrysh:
Won't fix. This is already on ocp-global-demo-wip as 83304cb3, so there is nothing left to merge here.

Issue #384, opened 2026-07-30
Nothing in eRoT state survives a power cycle


chrysh:
Retry budgets, isolation state, and Locked itself are RAM-only — a power cycle is a lockdown escape. LatchLockdown delegates durability to a shell that doesn't exist yet. Ask what the persistence story is; it interacts with the pending-update record (#383), so they may share a mechanism.

cc @FerralCoder @rusty1968

FerralCoder:
I would like to consult our partners on how this is typically handled today.

- Does eRoT/PRoT save any persistent data on power cycle?
- Under what condition would PRoT be power cycled? I am assuming this is rare (i.e. not cycled along with the rest of the system such as when updating/recovery of downstream components like BMC or CPU etc.

Issue #383, opened 2026-07-30
Updates: crash-loop handling and pending-update record


chrysh:
When we flash a new image and it verifies, you activate it immediately. What happens if it boots into a crash loop? I'd expect: boot the new image once, watch it actually come up, and only then commit to it — and if we lose power in the middle of an update, something written to flash has to remember where we were. Where does that live?

cc @FerralCoder @rusty1968

FerralCoder:
First, when to activate is a policy decision, so it will not necessarily be immediate.

When writing to a flash partition:

1. We never write to the active partition. If there is a power failure during update, the update can simply be restarted from be beginning after boot. This would be the responsibility of the outside entity that is driving the PLDM Type 5 update process. Since we did not touch the active partition, the device should be able to boot just as it was before.
2. Once writing of the update completes successfully, we mark the updated partition as active.
3. When the device next boots up, we must monitor the boot flow.
  - If we detect successful boot, update the inactive redundant partitions to match
  - if we detect boot failed:
    - put the device back into reset
    - enter recovery flow
      - find a valid inactive partition and and mark active
      - allow device to boot again
      - update the failed partition with the working copy again

chrysh:
Splitting this in two, because the answers live in different places.

The record is #478: one stored session for the eRoT's own update, holding the
state and the verified SVN together. The boot after a self-reset reads it and
knows whether the trial was ever armed, whether it ran, and whether it was
confirmed while the anti-rollback floor has not moved yet. Those three look the
same if you only store the SVN, and they need opposite actions.

Stopping a crash loop is the platform's job, not the record's. A timeout stops a
single bad boot: the deadline runs out, the orchestrator reverts and resets. It
cannot stop the loop, because a trial image that panics before the orchestrator
starts writes nothing, so the next reset boots the same image again.

Three ways to bound it:

- the bootloader clears the arming before it jumps into the trial image, so the
  trial gets one boot;
- a boot counter next to the active-partition record, decremented on each boot,
  falling back at zero (the Caliptra MCU partition table does this with
  `boot_count` and `status`);
- failing both, the trial image marks the attempt as used as its very first
  write, which still loops if it panics before that.

Which one we can use depends on whether the eRoT's boot selection is ours or
ROM-defined, which is still open.

PR #460, opened 2026-09-08
orchestrator: add TrialBoot capability trait

chrysh:
TL;DR: Trait-only PR. `TrialBoot` is the commit gate activation leaves open:
keep an image that was activated but not committed once its boot was judged,
or throw it away.

Part of OpenPRoT/openprot#377, 9elements/openprot#21.

## Summary

Activating an update picks the staged image as the next thing to boot and
stops there, whether the eRoT does that itself through `Updatable::activate`
or a PLDM firmware device does it for the eRoT. `Updatable`'s docs already
point at a `TrialBoot` capability as the other half; that trait did not
exist. This adds it to `orchestrator-capabilities`: `is_pending`, `confirm`,
`revert`.

`confirm` and `revert` take no arguments. A device has at most one trial open
at a time, the one the last activation armed, so there is nothing to name,
and which slot is which stays behind the seam as it does in `Updatable`.

## One contract, both implementations

The same trait covers a downstream device behind interposed flash and the
eRoT's own image. They differ in who judges the boot, not in what confirm or
revert means. A device that commits on its own, a PLDM firmware device
picking its own slot, has no `TrialBoot` on the eRoT side, the same split
`SvnFloor` already makes.

Three contract points:

- `is_pending` is answerable from stored state alone, because the image that
  confirms a self-update is not the image that armed it.
- The arming counts for the next boot only, but the record of the trial
  outlives it. An image that hangs, or an eRoT that loses power during the
  trial, boots the confirmed slot again with no call from anyone; the record
  stays open until `confirm` or `revert`, so an orchestrator that rebooted
  mid-update finds the trial again instead of losing it.
- `revert` also disarms. An orchestrator that drops an activation before
  resetting the device needs the next boot to run the confirmed slot even
  though the trial image never booted.

Both calls are safe to repeat: with no trial open they succeed and do
nothing, matching `SvnFloor::advance`.

`confirm` moves slot metadata only. Raising the anti-rollback floor is
`SvnFloor::advance` and happens later, on the UA's UpdateSecurityRevision
(DSP0267 1.3.0, command 0x22), which is what keeps `revert` able to fall back
to an image that still boots. See #464.

## Why not a method on `Updatable`

For self-update, staging and confirm run in different boots of different
images. The `Updatable` instance that staged the payload no longer exists
when `confirm` runs. A method on `Updatable` cannot model that.

Also considered and rejected: the consuming-session shape of #459 (a
session's compile-time guarantee cannot cross a power cycle, which is the
boundary this trait exists to span), and a constructor returning
`Option<Trial>` (models only cold-boot recovery, not the downstream happy
path where the trial opens mid-session).

## Tests

Seven host tests, both implementations driven through one generic
`apply_verdict<T: TrialBoot>` so the shared contract is what is under test: a
healthy downstream trial becomes confirmed, one nobody confirms falls back
yet stays open, revert before the trial boots disarms the next boot, the eRoT
confirms from the boot its trial started, an eRoT trial that never confirms
falls back on the next reset, a replayed resolution is a no-op, errors
surface through the generic seam.

## Not in this PR

The implementations. What backs the trial record durably is open
(OpenPRoT/openprot#383, #384): `services/storage` is still a stub, so the
eRoT-side impl has nowhere to keep the record yet.


services/orchestrator/capabilities/src/trial_boot.rs
```diff
@@ -0,0 +1,330 @@
...
+        fail: bool,
+    }
+
+    impl<'a> TrialRecord<'a> {
+        fn over(record: &'a mut SlotRecord) -> Self {
+            Self {
+                record,
+                fail: false,
+            }
+        }
```
leongross, line 198: Not too important for the test itself, but since tests often serve as examples for later implementation/integration, maybe the `From` trait fits better here, since it is a common API and imo fits here well.

chrysh: Fixed

rusty1968 (approved):

Issue #382, opened 2026-07-30
Orchestrator: golden partition fallback vs. policy


chrysh:
Does the orchestrator always boot into golden partition on fail, or does it take policy into account?

Context: https://github.com/OpenPRoT/openprot/pull/357/changes#diff-ed5cbc8c52e870e0a1d115475283d5916ded79625514a3d8b77cff48fb61cc3cR255

cc @FerralCoder @rusty1968

FerralCoder:
recovery flow is policy - there may be redundant A/B/C partitions to try before falling back to golden/recovery image. If redundant partitions are available, we would try those first. Falling back to golden/recovery image is last resort

chrysh:
@rusty1968 this is the policy question for that, concretely.

The core counts recovery attempts and caps them. Effect::RecoverComponent
carries `attempt`, and the driver is meant to pick a different source per
attempt: slot A on 0, slot B on 1, golden on 2. What should the driver do when
it runs out of sources before the core runs out of attempts?

Option one, return an error. It reaches the SM as EffectError and is treated
fail-closed, so a board with three sources latches Locked on attempt 3 even if
the cap is 5. The board's source count silently sets the real cap.

Option two, clamp to the last source and keep restoring from it until the
core's cap is spent. The cap stays meaningful, but the retries repeat a source
that already failed.

A board with one source never sees the difference, it just repeats the same
restore each attempt. The Recovery capability trait I am writing currently says
exhausted is an actuation error, which is option one. Which do you want?

rusty1968:
Addressed  at the SM level by #473 as a third option beyond the two originally proposed:
- not option one (return an error → EffectError → unconditional Locked, which as @chrysh noted lets a board's source count silently set a lower real cap than the core's, even for Isolable/Cascading components)
- not option two (clamp to the last source and burn remaining attempts repeating a restore that already failed)
- instead: the driver reports exhaustion as its own event, and the SM gates it through the same FailurePolicy machinery as the retry-cap-exhausted path (exhaust_recovery, shared between both) — Isolable/Cascading components are skipped, only Required locks the platform.

PR #463, opened 2026-09-10
orchestrator: add Recovery capability trait

chrysh:
The `Recovery` capability trait: rewrite one managed device's active image from
its board-configured recovery source. This is the seam
`Effect::RecoverComponent` needs, which currently falls through to
`Err(EffectError)` in the platform driver.

`Ok` means the mechanism completed, not that the image is good. Judging the
restored image stays with the verifier on the re-walk, so a restore can never
forge a verdict. Errors are actuation faults only and are treated fail-closed.

`restore` takes the `attempt` count straight from the effect, so an implementor
holding several sources can pick a different one each try (slot A on 0, slot B
on 1, golden on 2). A count kept by the device would drift, because it never
sees which attempt succeeded.

Trait only: no driver executor, no `BoardCapabilities` change, no board wiring.
Those come in a follow-up once the open question below is settled.

Open question, asked in #382: what should the driver do when it runs out of
sources before the core runs out of attempts? This PR says exhausted is an
actuation error, which fail-closes and lets the board's source count silently
set the real retry cap. The alternative is clamping to the last source.

Draft until #382 is answered.

Related: #48.

### DIFF
--- a/services/orchestrator/capabilities/BUILD.bazel
+++ b/services/orchestrator/capabilities/BUILD.bazel
@@ -11,6 +11,7 @@ rust_library(
         "src/evidence.rs",
         "src/lib.rs",
         "src/lockdown_latch.rs",
+        "src/recovery.rs",
         "src/svn_floor.rs",
         "src/updatable.rs",
     ],
--- a/services/orchestrator/capabilities/src/lib.rs
+++ b/services/orchestrator/capabilities/src/lib.rs
@@ -25,6 +25,10 @@
 //! `BootWatch` is the seam the orchestrator polls: one device's boot walk,
 //! erased of every device-specific type, answering with a `WalkVerdict`.
 //!
+//! `Recovery` is the restore capability: rewrite one device's active image
+//! from its board-configured recovery source, mechanism unnamed, source
+//! chosen per attempt.
+//!
 //! `LockdownLatch` is the terminal capability: latch the platform into its safe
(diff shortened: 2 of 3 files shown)


rusty1968:
The questions are answered by https://github.com/OpenPRoT/openprot/pull/473

rusty1968:
`Recovery::restore()` returns a single opaque `Result<(), Self::Error>`, and the doc note above it says exhaustion is "an actuation error like any other." That conflicts with the contract #473 just added on `Platform::execute` (`services/orchestrator/sm/src/lib.rs`): a driver must distinguish "no recovery source left" (→ `Event::RecoveryUnavailable`, gated per `FailurePolicy`) from a genuine actuation fault (→ `EffectError`, unconditional lockdown). As written, whoever implements the `RecoverComponent` executor has no way to tell those apart from `Self::Error` alone.

Suggest extending the contract so an implementor can signal exhaustion distinctly, e.g.:

```rust
enum RestoreOutcome {
    Restored,
    SourcesExhausted,
}

fn restore(&mut self, attempt: u8) -> Result<RestoreOutcome, Self::Error>;
```

so `Self::Error` stays reserved for real actuation faults, and `SourcesExhausted` maps directly to `Event::RecoveryUnavailable`.

rusty1968 (changes requested):
See comments.

rusty1968 (approved):

PR #468, opened 2026-09-11
Expanding firmware_update to pull data.  Added inventory test.

CourtneyDrant:
Added a FW payload to firmware_update tests to stress FD as an initiator.  Added a test that explicitly focuses on Inventory commands.  Inventory commands do not invoke the state machine and are limted to GetFirmwareParameters and QueryDeviceIdentifiers.


services/pldm/tests/firmware_update_host.rs
```diff
@@ -218,10 +232,14 @@ fn serve_ua_fw_request<S: Sender, const N: usize>(
...
                 RequestFirmwareDataRequest::decode(payload).expect("decode RequestFirmwareData");
+            let offset = fw_req.offset as usize;
             let length = fw_req.length as usize;
             assert!(length <= MAX_TRANSFER_SIZE, "requested chunk exceeds MTU");
-            let data = [0xA5u8; MAX_TRANSFER_SIZE];
-            let resp_msg = RequestFirmwareDataResponse::new(instance_id, success, &data[..length]);
+            let end = offset.checked_add(length).expect("firmware range overflow");
+            let data = component_image
+                .get(offset..end)
+                .expect("requested range should be within the component image");
```
chrysh, line 241: nit: consider clamping to the image length and zero-padding the remainder instead of panicking, matching DSP0267's allowance for past-end requests. Not needed today since the FD always requests exact remaining bytes, but would save a surprise if a padding test lands later.

services/pldm/tests/inventory_host.rs
```diff
@@ -0,0 +1,302 @@
+// Licensed under the Apache-2.0 license
+// SPDX-License-Identifier: Apache-2.0
+
+//! End-to-end host test for PLDM firmware-update inventory commands.
+//!
+//! The Update Agent queries a Firmware Device over in-memory MCTP and verifies
+//! the type 5 device identifiers and firmware parameters supplied by `FdOps`.
+
```
chrysh: nit: "a initiator" → "an initiator"

services/pldm/tests/inventory_host.rs
```diff
@@ -0,0 +1,302 @@
...
+    };
+    let fd_server: RefCell<Server<_, 16>> = RefCell::new(Server::new(Eid(FD_EID), 0, fd_sender));
+
+    // Create the DirectClientWithPump for the responder role, which handles requests from the UA to the FD.
+    let responder_client = DirectClientWithPump::new(&fd_server, || {
+        transfer(&ua_to_fd_packets, &mut fd_server.borrow_mut());
+        ua_to_fd_packets.borrow_mut().clear();
+    });
+
+    // Inventory Commands do not invoke the FW FSM and do not invoke FD as a initiator.  Therefore the closure is empty.
```
chrysh: nit: "a initiator" → "an initiator"

chrysh (approved):
Only two nits, but otherwise it's ready to merge

services/pldm/tests/inventory_host.rs
```diff
@@ -0,0 +1,302 @@
...
+        0,
+        &active_version,
+        &pending_version,
+        ComponentActivationMethods(0x0009),
+        FirmwareDeviceCapability(0x0000_0003),
+    );
+    let firmware_parameters = FirmwareParameters::new(
+        FirmwareDeviceCapability(0x0000_0003),
+        1,
+        &active_image_set_version,
```
chrysh, line 178: nit: 0x0009 and 0x0000_0003 are magic numbers. Named constants or bitwise OR of the individual flags (if pldm-common exposes them) would make the test easier to read. Failing that, a comment saying which bits are set (AC power cycle | system reboot, update mode restriction | partial updates) would help.

CourtneyDrant: I understand and agree but these values are throwaway, they mean nothing.  We simply need them to be populated.

PR #471, opened 2026-09-17
orchestrator: Add CheckpointWalk, the concrete BootWatch

chrysh:
## Summary

- Adds `CheckpointWalk<R, G>`, the concrete `BootWatch` that walks a device's
  `BootCheckpoint`s in declaration order, polling an `EvidenceReader` at each
  step and judging per-checkpoint windows against caller-injected `now_millis`.
- Refactors QEMU runtime itest scenarios 1-2 to use `CheckpointWalk` directly,
  separating walk judgment from `BootWatchdogs` scheduling.
- Adds a GPIO wiring example to `BootCheckpoint`'s doc comment in the config
  crate.

Read errors are silence (`Booting`), lapsed windows are timeouts with no
last-chance read. 17 host unit tests cover happy path, timeout, device faults,
read errors, arm/re-arm, deadline arithmetic, and construction rejection.

## Test plan

- [x] `bazel test //services/orchestrator/adapters/walk:orchestrator_checkpoint_walk_test` (17 host tests)
- [x] `bazel test //services/orchestrator/config:orchestrator_config_test`
- [x] `bazel test --config=virt_ast10x0 //target/ast10x0/tests/orchestrator/runtime:runtime_test` (5 QEMU scenarios)
- [x] `bazel test --config=virt_ast10x0 //target/ast10x0/tests/orchestrator/runtime:no_panics_test`


rusty1968:
@JesseMelon  eyes on this.

rusty1968 (approved):

services/orchestrator/adapters/walk/src/walk.rs
```diff
@@ -0,0 +1,525 @@
...
+}
+
+impl<R: EvidenceReader<P>, P> BootWatch for CheckpointWalk<R, P> {
+    fn arm(&mut self) {
+        self.phase = Phase::Armed;
+    }
+
+    fn poll(&mut self, now_millis: u64) -> WalkVerdict {
+        if let Phase::Armed = self.phase {
+            let timeout_millis = self.checkpoints[0].timeout().as_millis() as u64;
```
rusty1968, line 65: BootCheckpoint::new has one check: "the timeout can't be zero." Pass that check, and you'd assume you're fine.

But this code that actually uses the timeout secretly rounds it down to whole milliseconds first. So "not zero" isn't really the rule the code needs — the real rule is "at least 1 millisecond." Nobody checks for that.

I'd tighten `BootCheckpoint::new`'s assert to reject sub-millisecond durations

services/orchestrator/adapters/walk/src/walk.rs
```diff
@@ -0,0 +1,525 @@
...
+}
+
+impl<R, P> CheckpointWalk<R, P> {
+    /// Binds a reader and its checkpoint list into a walk. The checkpoints
+    /// are walked in declaration order; each one's probe is resolved by
+    /// the reader.
+    ///
+    /// # Panics
+    ///
+    /// Panics if `checkpoints` is empty.
```
rusty1968, line 40: Why is CheckpointWalk::new the one constructor in this family that validates at runtime instead of compile time — should it require its input already come from a DeviceConfig (compile-time validated) rather than a raw &'static[BootCheckpoint<P>] slice?"

services/orchestrator/adapters/walk/src/walk.rs
```diff
@@ -0,0 +1,525 @@
...
+    pub fn new(reader: R, checkpoints: &'static [BootCheckpoint<P>]) -> Self {
```
rusty1968, line 41:
```rust
   /// checkpoints were already checked non-empty and uniquely-named at
    /// build time by `DeviceConfig::new` — nothing left to assert here.
    pub fn new<S>(reader: R, device: &'static DeviceConfig<S, P>) -> Self {
        Self {
            reader,
            checkpoints: device.checkpoints(),
            phase: Phase::Idle,
        }
    }
```

rusty1968:
@chrysh , you can fold the feedback into the folow-up PR that addresses https://github.com/OpenPRoT/openprot/issues/477

